PR #717, opened 2012-12-28
Add ability to use CSS class for label of item in the collection for radio and checkbox.

yurikoval:


flyfy1:
I have similar needs -- our own style guide requires me to apply a class on the item level of a collection.

Regarding with the *currently supported block version*, would anyone point me to the reference for that? Thanks! @carlosantoniodasilva @rafaelfranca

carlosantoniodasilva:
@flyfy1 is this what you're looking for? https://github.com/plataformatec/simple_form/blob/f98b158a605f9a76dd0a4f60ae8549a3c5f863a5/lib/simple_form/form_builder.rb#L345-L351

flyfy1:
@carlosantoniodasilva I don't get what u mean... where is it appending the class into options?

BTW... I think I've probably solved the problem via MonkeyPatching 😉

carlosantoniodasilva:
@flyfy1 Ah gotcha!

You should be able to pass options to the `label` and `radio_button` methods there, for example:

```ruby
b.label(class: 'zomg-label') { b.radio_button(class: 'lol-radio') + b.text }
```

This is [documented in Rails](https://github.com/rails/rails/blob/20c91119903f70eb19aed33fe78417789dbf070f/actionview/lib/action_view/helpers/form_options_helper.rb#L638-L642) now that the bulk of these collection methods has been moved to it. I hope that helps!

PR #786, opened 2013-04-17
Consider whether the user explicitly specifies a label when applying global custom formatting

timscott:
Let's say I want to titleize my generated labels.  I can do it:

``` ruby
config.label_text = lambda { |label, required| "#{label.titleize}" }
```

But then I can't have a non titled label, even if I specify it.  This:

``` ruby
<%= f.input :distributor, label: 'Who provides your cable or satellite service?' %>
```

Will generate this label: `Who Provides Your Cable Or Satellite Service?`

Yeah, ugly.  It's not what I want, and there's no way to handle it (except maybe with undesirable options like #human_attribute_name or by replacing a major part of the framework.)

This PR is a simple solution that does not break existing code.

### DIFF
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -66,7 +66,7 @@ module SimpleForm
 
   # How the label text should be generated altogether with the required text.
   mattr_accessor :label_text
-  @@label_text = lambda { |label, required| "#{required} #{label}" }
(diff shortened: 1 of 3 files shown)


brian-vogogo:
Just used this feature. Thanks everyone!

PR #835, opened 2013-06-29
Object mapper abstraction support.

eric1234:
Initial support for ActiveRecord and MongoMapper. Provides obvious place to
add support for addition object mappers and to enhance additional support
needed on ActiveRecord and MongoMapper.

Browing through the MANY tickets regarding adding an object mapper abstraction
it seems the preferred method is to try to let ActiveModel handle most of the
abstraction and just extend the various object matters where ActiveModel is
not abstracting enough. This way most of the SimpleForm code does not care
what object mapper is being used.

Is seems to me there are two classes of methods that we need to extend the
object mappers with.
1. Methods that return simple data. A great example of this is the
   "readonly_attributes" method. SimpleForm uses this method that ActiveRecord
   provides but an object mapper like MongoMapper doesn't have the concept of
   read only attributes. So in this case we take the simple approach and simply
   mimic the ActiveRecord method by returning an array (just like ActiveRecord)
   which is empty (since the concept is not supported).
2. Methods that return complex data. A good example of this is
   "reflect_on_association". This returns an ActiveRecord-specific data
   structure. So in this case I made all data mappers (AR and MM) return a
   object mapper independent data structure (SimpleForm::Association). Because
   I am returning an SimpleForm specific data structure I have prefixed these
   methods with "simple_form_" so we reduce poluting the object mapper's
   namespace.

This second style method is basically what the testing is already doing. We
are just moving those object mapper independent data structures into the code
instead of having them only used in testing.

I have no doubt that this patch is missing stuff we will need down the road.
I am only using pieces of SimpleForm for my project so there may be areas where
we need more abstraction methods added to the object mappers. Obviously others
are also interested in other object mappers (DataMapper, Mongoid, etc). I feel
like that can come with time by additional patches. Right now we just need a
place to start. An obvious place to put these abstraction methods. Otherwise
we will continue to have dozens of issues where everybody want it but it cannot
get off the ground.

Regarding testing, I did add testing for the two object mapper independent
objects I added (Column and Association). But I was unsure the approach to take
regarding testing the abstraction methods that we add to the object mappers.
I was unsure if we wanted to required the tests to have all those database
types available. Or maybe have tests but skip ones not available. In general
the abstraction methods are simple. Just copying data from the object mapper
specific data structures to the object mapper independent data structures.

If you have suggestions on how to improve this patch I am open and ready to
follow suggestions. But I want to ensure we don't let perfection get in the
way of progress. For me it is ok if we have to refactor our approach as we add
more object mappers in the future. Sorry to be long winded. :)

### DIFF
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -1,6 +1,7 @@
 require 'action_view'
 require 'simple_form/action_view_extensions/form_helper'
 require 'simple_form/action_view_extensions/builder'
+require 'simple_form/oms'
 require 'active_support/core_ext/hash/slice'
 require 'active_support/core_ext/hash/except'
 require 'active_support/core_ext/hash/reverse_merge'
@@ -10,6 +11,8 @@ module SimpleForm
 
   autoload :Helpers
   autoload :Wrappers
+  autoload :Column
+  autoload :Association
 
   eager_autoload do
     autoload :Components
--- a/lib/simple_form/association.rb
+++ b/lib/simple_form/association.rb
@@ -1,0 +1,27 @@
+# An Association object represents the specs of an relationship as
+# defined by the Object Mapper in an Object Mapper independent mannor.
+# A Object Mapper independent version of
+# ActiveRecord::Associations::Association
+class SimpleForm::Association
+
+  # Should be :belongs_to, :has_one or :has_many
+  attr_accessor :macro
+
+  # The type of object this association relates to
+  attr_accessor :klass
+
+  # The name of this association
+  attr_accessor :name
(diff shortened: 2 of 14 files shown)


feliperenan:
Do we have plans to support it?

cc @rafaelfranca

feliperenan:
I'm gonna close this PR since it hasn't recently activity and I don't know if we have plans to support this feature.

Thanks

rafaelfranca:
I'm fine with this if people wants to do the work to support others ORM, but definitely don't want to maintain things related to mongoid

PR #997, opened 2014-03-11
Make possible to use the Wrappers API to define attributes for the components

rafaelfranca:
(This is a alternative solution to https://github.com/plataformatec/simple_form/pull/622)
## Motivation

To integrate Simple Form with Bootstrap 3 we need to make possible to define classes to input only in some wrappers.

We added `input_class` some time ago but that [broke some components like checkboxes](https://github.com/rafaelfranca/simple_form-bootstrap/commit/7d76738a01d647d77447cc38afd6786d4f69af83) (see https://github.com/rafaelfranca/simple_form-bootstrap/pull/28).

Instead of changing a global class to be applicated in every input we allow users to choose in which wrappers they want to define a class.
## How to use

Now users can define attributes for the components in the component definition of the Wrappers API.

``` ruby
config.wrappers tag: 'div', class: 'control-group', error_class: 'error' do |b|
    b.use :label, class: 'form-label'
    b.use :input, class: 'form-input'
end
```

The `class: 'form-label'` option will be used every time a label component is rendered.
## Implementation

To make this possible I had to change who the Wrappers API is defined and how components are rendered.
### Wrapper API storage change

Before the work on this branch, the API components were stored in a mix of Simple Form classes and symbols.

simple_form_wrapper_before

This added some conditionals in our rendering code and made harder to store state in our components. This create a lot of code in Simple Form that rely in the fact that Ruby hashes are mutables and we have a global state. The implementation in #622 had this problem were the changes in a global hash make the [hint and errors classes to getting duplicated](https://github.com/plataformatec/simple_form/pull/622#issuecomment-28395418).

I choose to avoid completely this problem storing the attributes in a class and making it implement the same protocol as `SimpleForm::Wrappers::Many` and `SimpleForm::Wrapper::Single`. This also make possible to remove some conditionals to deal with Symbols in the components tree.

simple_form_wrapper_after
### Input rendering

Before this work, when rendering the input, there was no way to pass information about the context where it is being rendered. I changed the implementation to always pass the wrapper options for the component rendering method itself.

To understand better lets look how a custom input was implemented before:

``` ruby
class CustomInput < SimpleForm::Inputs::Base
  def input
    @builder.text_field(attribute_name, input_html_options)
  end
end
```

So when rendering a wrapper Simple Form will call the method `input` of that class. As you can see we don't have a clean way to get information about the wrapper that is being rendered inside that method.

Now, Simple Form will pass an argument `wrapper_options` that hold the options defined in the Wrappers API. So, for our previous example, `wrapper_options` will contain `class: 'form-input'`.

So, our custom input will look like:

``` ruby
class CustomInput < SimpleForm::Inputs::Base
  def input(wrapper_options)
    @builder.text_field(attribute_name, input_html_options)
  end
end
```

Now we can use a new helper `merge_wrapper_options` to merge the options present in the `form.input` call and the options defined in the wrapper.

Finally, our custom input will look like:

``` ruby
class CustomInput < SimpleForm::Inputs::Base
  def input(wrapper_options)
    @builder.text_field(attribute_name, merge_wrapper_options(input_html_options, wrapper_options))
  end
end
```
## Deprecations

The custom input/custom components methods now have to accept one argument with the wrapper options. The inputs/components defined using the old API will work but with a deprecation message.

These examples are deprecated:

``` ruby
class CustomInput < SimpleForm::Inputs::Base
  def input
  end

  def label
  end

  def hint
  end

  def label_input
  end

  def error
  end

  def placeholder
  end
end
```

To fix the deprecation message use:

``` ruby
class CustomInput < SimpleForm::Inputs::Base
  def input(wrapper_options)
  end

  def label(wrapper_options)
  end

  def hint(wrapper_options)
  end

  def label_input(wrapper_options)
  end

  def error(wrapper_options)
  end

  def placeholder(wrapper_options)
  end
end
```
## Next steps
- [x] Add proper documentation
- [x] Test this new feature writing the wrappers for Bootstrap 3

Closes #622

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,6 @@ gemfile:
   - gemfiles/Gemfile.rails-4-0-stable
   - gemfiles/Gemfile.rails-4-1-stable
   - Gemfile
-matrix:
-  allow_failures:
-    - gemfile: Gemfile
-    - gemfile: gemfiles/Gemfile.rails-4-1-stable
 notifications:
   email: false
   campfire:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   * Add `include_default_input_wrapper_class` config [@luizcosta](https://github.com/luizcosta)
   * Map `datetime`, `date` and `time` input types to their respective HTML5 input tags
   when the `:html5` is set to `true` [@volmer](https://github.com/volmer)
+  * Add `boolean_label_class` config.
+  * Add `:html` option to include additional attributes on custom wrappers [remofritzsche](https://github.com/remofritzsche) and [@ulissesalmeida](https://github.com/ulissesalmeida)
   * Make possible to use the Wrappers API to define attributes for the components.
   See https://github.com/plataformatec/simple_form/pull/997 for more information.
 
--- a/README.md
+++ b/README.md
@@ -346,14 +346,14 @@ used to retrieve label/value attributes for the `option` tags. Besides that, you
 ### Priority
 
 **Simple Form** also supports `:time_zone` and `:country`. When using such helpers, you can give
-`:priority` as option to select which time zones and/or countries should be given higher priority:
+`:priority` as an option to select which time zones and/or countries should be given higher priority:
 
 ```ruby
 f.input :residence_country, priority: [ "Brazil" ]
 f.input :time_zone, priority: /US/
 ```
 
-Those values can also be configured with a default value to be used site use through the
+Those values can also be configured with a default value to be used on the site through the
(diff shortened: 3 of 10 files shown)


AndreLGava:
Hello, can anyone help me I'm still learning rails and was trying to create a input to accept datepicker-bootstrap, I found this tutorial https://github.com/plataformatec/simple_form/wiki/Custom-inputs-examples, I createad as it was declared and worked for a form but when I load another don't, in my terminal I found the reason, that must be this issue and this deprecated, can you help me about this broblem?

Issue #492, opened 2012-03-02
Add space to checkboxes and radio buttons


ryanb:
When using a `collection` with checkboxes or radio buttons, there is no space between the checkbox/radio button and the label. But there is a space for a boolean attribute checkbox. I think a space should be added to the collections to make it consistent with boolean attributes.

PR #1008, opened 2014-03-28
Put a whitespace before the inline_label if it is present

rafaelfranca:
To align with the Bootstrap interface we need to put an space before the inline label:

screen shot 2014-03-28 at 18 43 30

So when we call:

``` erb
<%= f.input :checkbox, as: :boolean, inline_label: 'Remember me' %>
```

The label will be placed correctly.

I'm not sure if we should do this by default or require the users to put the whitespace in the label text like this:

``` erb
<%= f.input :checkbox, as: :boolean, inline_label: ' Remember me' %>
```

cc @carlosantoniodasilva


willkoehler:
There is now a inconsistency between the label for a boolean attribute checkbox and the labels in a collection of checkboxes. The labels in a collection of checkboxes don't have a leading space. Basically a repeat of this issue https://github.com/plataformatec/simple_form/issues/492.

Can you add a leading space to the labels in a collection of checkboxes to make the layout consistent? I can open a new issue if that would help.

rafaelfranca:
Open a new issue please

PR #1094, opened 2014-06-29
Remove call to column_for_attribute as it's depreciated in rails master

ghost:
The title says it all, `column_for_attribute` is now depreciated to this pull requests removes the call and replaces it with a call to `@object.class.columns_hash`.


akostadinov:
The method is not deprecated, only the return value of `nil`. Just for completeness.

PR #1109, opened 2014-07-23
Allow overriding of default configured form_class

rmm5t:
This is especially useful with bootstrap forms where you might want to switch to a `form-inline` layout even though the default is a `form-horizontal`.

e.g.

``` ruby
config.form_class = "form-horizontal" # configured default
```

``` ruby
simple_form_for @user, form_class: "form-inline"
```

I recognize that this was (somewhat) already proposed and closed in #657 and #639, but I feel like this new implementation is more straight forward, cleaner, and more discoverable without passing around non-standard html attributes. It's merely a config override.

Just throwing this out there.


bquorning:
I was surprised to see that there’s no link between this PR and the issue #1217. Well, now there is :-)

PR #1201, opened 2015-02-04
Update legacy v2.2 branch

johnnyshields:
I was using the v2.2 branch via `:git` reference, and seems it has fallen behind v2.1 branch. This PR adds in the fixes from v2.1 to v2.2. Please consider releasing v2.2 if possible in order to tie up the loose ends.


rafaelfranca:
I don't think we are going to release 2.2. I started this branch as experiment to see if we could backport some of the 3.0 features to 2.x release but it was not worth.

johnnyshields:
OK, will close PR then.

PR #1211, opened 2015-02-24
Can't set collection for configured custom value input

strangeworks:
Hello everyone! I have the following configuration for my custom form fields:
```ruby
config.input_mappings = {/bodytype/ => :check_boxes}
```
And would like to provide my custom collection generation helper for this kind of input, like this:
```ruby
f.input :bodytype, :collection => bodytype_collection
````
But it doesn't work for me because of explicit return in `default_input_type` method, my quick and dirty solution was simple, bump custom time detection statements top. Is that an issue, or it is impossible to use :collection option by design?

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -508,8 +508,8 @@ module SimpleForm
     # collection is given.
     def default_input_type(attribute_name, column, options)
       return options[:as].to_sym if options[:as]
-      return :select             if options[:collection]
       custom_type = find_custom_type(attribute_name.to_s) and return custom_type
+      return :select             if options[:collection]
 
       input_type = column.try(:type)
       case input_type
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/simple_form/form_builder.rb [hunk 1]
```diff
@@ -507,9 +507,10 @@ def find_input(attribute_name, options = {}, &block)
     # default alwayls fallback to the user :as option, or to a :select when a
     # collection is given.
     def default_input_type(attribute_name, column, options)
+      custom_type = find_custom_type(attribute_name.to_s) and return custom_type
+
       return options[:as].to_sym if options[:as]
```
carlosantoniodasilva: I think we should have it after this `as` condition, because that'd allow you to override custom type on a per-input basis, right?

carlosantoniodasilva:
We'd love if you could write a test for it too. Thanks!

lib/simple_form/form_builder.rb
strangeworks: Yep, it can cause an issue with overriding using as, thanks for notice!

strangeworks:
Done, waiting for feedback, thanks!

PR #1215, opened 2015-03-05
Update readme, add usage for namespaced model [ci skip]

anilmaurya:


nashby:
@anilmaurya hey, thank you for the pull request but that's how rails's `form_for` works for not sure we need to duplicate it in the README: http://guides.rubyonrails.org/form_helpers.html#dealing-with-namespaces

Issue #1114, opened 2014-07-31
Why is `f.button :button` a thing?


rmm5t:
Sorry. I just ran into this, and I'm probably missing some history, but the fact that the signature of `f.button` with simple_form is different than the core `f.button` method in Rails is a bit confusing. 

https://github.com/plataformatec/simple_form/blob/ed3de1c2b566a962697529e824953f6b280e2867/lib/simple_form/form_builder.rb#L193-L213

I understand that this was implemented to support `SimpleForm.button_class` config option, but wouldn't something like the following implementation be more direct/correct/?

``` ruby
alias_method :button_without_simple_form, :button
def button(value = nil, *args, &block)
  options = args.extract_options!.dup
  options[:class] = [SimpleForm.button_class, options[:class]].compact
  args << options
  button_without_simple_form(value, *args, &block)
end

alias_method :submit_without_simple_form, :submit
def submit(value = nil, *args)
  options = args.extract_options!.dup
  options[:class] = [SimpleForm.button_class, options[:class]].compact
  args << options
  submit_without_simple_form(value, *args)
end
```

Something like this would allow me to use the same `f.button` syntax in Rails with or without SimpleForm:

``` haml
  = f.button "Save Widget"
```

Instead of what is currently necessary while using SimpleForm:

``` haml
  = f.button :button, "Save Widget"
```

The latter just feels awkward.

Questions/Comments: 
- **Is the SimpleForm team willing to accept a pull-request that changes this?**
- **Ugh regarding backwards compatibility concerns here. I suppose this won't make it for a 3.x release.**

ryanhageman:
This was resolved, right?

PR #1218, opened 2015-03-09
Update README with a button element example

benlinton:


nashby:
@benlinton thanks!

Just for the reference why it's working like this: #1114

PR #1221, opened 2015-03-13
Update README.md

aried3r:
I was puzzled by this behaviour and found nothing in the README. Only after I dug around in the issues, it became clear to me what to do and I've seen quite a few issues created around this.

So I suggest placing this in the README. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ Also, if you want to use the country select, you will need the
 gem 'country_select'
 ```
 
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
Seems good, thanks.

carlosantoniodasilva:
I've moved things a bit in the readme with a new country select section, to make things clearer: fd4b32ba390e698b667cf45c49c228f0eed8ba1d. Thanks :heart:

aried3r:
Makes sense, guess that's what I was really looking for ;)

PR #1225, opened 2015-03-19
Update docs regarding item_wrapper_tag

### LINKED ISSUE #1224: Item wrapper tag not behaving as documented
Configuration option item_wrapper_tag is always applied. The [README](https://github.com/plataformatec/simple_form/commit/6e35bec1dca35a768073c70c63393c4cf2716929) says item_wrapper_tag is not used when boolean_style is set to :nested. The reasons for this behaviour  are [discussed here](https://github.com/plataformatec/simple_form/issues/685).

This behaviour changed when upgrading from SimpleForm 2.x to 3.0. Consequently if using default configuration when upgrading, all items of collection of radio/check boxes are now wrapped with additional span element. This is easily remedied with setting item_wrapper_tag to nil.

I think we can update the README to reflect the new (and better) behaviour.

This is really a minor issue in this greatly useful piece of software - keep up the good work :+1:

jurglic:
Item_wrapper_tag configuration option in 3.x behaves independently, without any limitations imposed by nested boolean_style as mentioned in the documentation.

Update readme and a comment inside the initializer.

Closes #1224.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,6 @@ end
 ```
 
 For check boxes and radio buttons you can remove the label changing `boolean_style` from default value `:nested` to `:inline`.
-Also, `item_wrapper_tag` will not work when `boolean_style` is set to `:nested`.
 
 Example:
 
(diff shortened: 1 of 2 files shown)


nashby:
@jurglic thank you!

PR #1226, opened 2015-03-21
Fix method reflection for ruby 2.1

badosu:
Fixes warning:

        /home/badosu/.gem/ruby/2.2.1/gems/simple_form-2.1.2/lib/simple\
        _form/inputs/boolean_input.rb:37:
        warning: circular argument reference - unchecked_value


badosu:
This build is failing on Ruby < 1.9.3, but that's because it's not supported on the gem `i18n`. So that should not be a problem caused by this PR.

carlosantoniodasilva:
Thanks.

badosu:
Thanks!

badosu:
Just a question, do you still release 2.1.x? What's the plan?

rafaelfranca:
Depending on the change yes, we may release a new 2.1 version. 2.2 was just an experiment but I don't plan to release a version from it.

badosu:
Thanks for the info.

PR #1204, opened 2015-02-10
correction bootstrap datetime field, issue #1186 

eltonchrls:
correction to issue #1186

```ruby
config.wrappers :horizontal_select_date, tag: 'div', class: 'form-group', error_class: 'has-error' do |b|
    b.use :html5
    b.optional :readonly
    b.use :label, class: 'control-label'
    b.wrapper tag: 'div', class: 'form-inline' do |ba|
      ba.use :input, class: 'form-control'
      ba.use :error, wrap_with: { tag: 'span', class: 'help-block' }
      ba.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
    end
  end
```
in html
```ruby
<%= f.input :date, wrapper: :horizontal_select_date %>
```

PR #1236, opened 2015-03-31
Apply nice format for datetime, date and time on bootstrap.

### LINKED ISSUE #1186: Datetime looks weird in rails 4.1 and bootstrap 3
I installed simple_form with default bootstrap cofiguration: <code> rails generate simple_form:install --bootstrap </code>

And I have something like this in my code
<code><%= f.input :start_at, as: :datetime %> </code>

But the result is:
screen shot 2015-01-05 at 9 06 15 pm

Which looks weird, they are supposed to be in the same line right ?  however, bootstrap2 was fine.

ulissesalmeida:
Thanks for the PR https://github.com/plataformatec/simple_form/pull/1204 we now can have a nice style
for our datetime fields. I've just applied that format for date and time.

It closes #1186 

Now looks ok:

screen shot 2015-03-31 at 8 20 11 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
   * `date/time/datetime` inputs now correctly generate the label `for` attribute when
   HTML5 compatibility is explicitly enabled. [@ericsullivan](https://github.com/ericsullivan)
 
+  * The boostrap datetime, date and time now has a nice format on bootstrap.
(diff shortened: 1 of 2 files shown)


rafaelfranca:
:shipit:

carlosantoniodasilva:
:green_heart::yellow_heart::heart::blue_heart::purple_heart:

PR #1239, opened 2015-04-09
minor update to fix alignment

ardianys:
in bootstrap every field input is inside div with grid definition in it's class.

### DIFF
--- a/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
+++ b/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
@@ -126,7 +126,7 @@ SimpleForm.setup do |config|
     b.use :html5
     b.optional :readonly
     b.use :label, class: 'control-label'
-    b.wrapper tag: 'div', class: 'form-inline' do |ba|
+    b.wrapper tag: 'div', class: 'form-inline col-sm-10' do |ba|
       ba.use :input, class: 'form-control'
       ba.use :error, wrap_with: { tag: 'span', class: 'help-block' }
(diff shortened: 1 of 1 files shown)


lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -126,7 +126,7 @@
     b.use :html5
     b.optional :readonly
     b.use :label, class: 'control-label'
-    b.wrapper tag: 'div', class: 'form-inline' do |ba|
+    b.wrapper tag: 'div', class: 'form-inline col-sm-10' do |ba|
```
carlosantoniodasilva: This seems to only be necessary in the horizontal form, right? How would that affect a non horizontal form?

Also, other horizontal forms tend to use `col-sm-3` in the label and `col-sm-9` in the field, any particular difference?

Thanks.

ardianys: Yeah, it's just affect horizontal form I think.

I use col-sm-10 since I saw it on the other form element.

If this minor addition shouldn't located here, we should add new option to
change the "<div inline form>" class

Thanks.

rafaelfranca:
@herminiotorres this need to be tested in the example application.

feliperenan:
@rafaelfranca Are you considering merge it?

rafaelfranca:
If this works, yes

feliperenan:
@ardianys Can you provide to us a example application?

feliperenan:
Closing in favor https://github.com/plataformatec/simple_form/pull/1569

PR #1247, opened 2015-05-03
Make it possible to override required value that was set in the wrapper.

### LINKED ISSUE #1178: when specifying `required` html5 attribute in wrapper, override at input declaration requires opposite logic
I have a wrapper configured like so:

```ruby
  config.wrappers :floating_label, tag: 'div', class: 'FormGroup', error_class: 'form--error' do |b|
    b.use :html5
    b.use :placeholder
    b.use :input, autofocus: false, required: true
    b.use :label
    b.use :error, wrap_with: { tag: :p, class: 'form__error' }
  end
```

When I use this wrapper in simple form, in order to override the html5 validation, I need to use `true`

```ruby
  = simple_form_for @family do |f|
    = f.input :avatar, required: true
```

rather than what I would expect, which is to set it to false

```ruby
  = simple_form_for @family do |f|
    = f.input :avatar, required: false
```

Even though I'm not super familiar with development on the gem, If you point me to the proper specification I think I could write a failing test, and maybe fix.

nashby:
fixes #1178 

@jedschneider hey, I fixed you test and added it to this pull request.

### DIFF
--- a/test/inputs/required_test.rb
+++ b/test/inputs/required_test.rb
@@ -118,6 +118,7 @@ class RequiredTest < ActionView::TestCase
           concat f.input :name, required: false
         end
         assert_no_select 'input[type=text][required]'
+        assert_no_select 'input[type=text][aria-required]'
       end
     end
   end
--- a/test/inputs/required_test.rb
+++ b/test/inputs/required_test.rb
@@ -110,4 +110,15 @@ class RequiredTest < ActionView::TestCase
     assert_no_select 'input[required]'
     assert_select 'input.optional#validating_user_phone_number'
   end
+
+  test 'builder input does not generate required html attribute when option is set to false when it is set to true in wrapper' do
+    swap SimpleForm, browser_validations: true do
+      swap_wrapper :default, self.custom_wrapper_with_required_input do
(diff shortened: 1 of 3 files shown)


rafaelfranca:
:shipit:

test/inputs/required_test.rb
```diff
@@ -110,4 +110,15 @@ class RequiredTest < ActionView::TestCase
...
     assert_select 'input.optional#validating_user_phone_number'
   end
+
+  test 'builder input does not generate required html attribute when option is set to false when it is set to true in wrapper' do
+    swap SimpleForm, browser_validations: true do
+      swap_wrapper :default, self.custom_wrapper_with_required_input do
+        with_concat_form_for(@user) do |f|
+          concat f.input :name, required: false
+        end
+        assert_no_select 'input[type=text][required]'
```
carlosantoniodasilva: Should we check for `aria-required` as well?

carlosantoniodasilva:
:+1:

test/inputs/required_test.rb
nashby: Good catch, thanks! Added.

PR #1248, opened 2015-05-04
Bug/fix circular argument reference

srpouyet:
Hi,

We're using Simple Form 2.1 with a legacy Rails 3.2 app. Rails compatibility with Ruby 2.2 was fixed in the 3-2-stable branch, but there was a 'circular reference warning' with the current Simple Form 2.1 release. 

Unfortunately we can't go to Simple Form 3+ since it's not compatible with Rails 3.2 so I fixed the circular reference (I created the bug fix branch from the 'v2.1' tag).

I'm getting a lot of warnings running the tests with Ruby 2.2.1 but it looks like they're still passing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,5 +8,6 @@ gem 'activemodel', '~> 3.2.0'
 gem 'actionpack', '~> 3.2.0'
 gem 'rake'
 gem 'rdoc'
+gem 'test-unit'
 gem 'mocha', '~> 0.13.0', :require => false
 gem 'tzinfo'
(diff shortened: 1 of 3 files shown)


carlosantoniodasilva:
Thanks for your PR, this has been fixed by #1226 and merged into v2.1 branch. You should be able to bundle from that:

```ruby
gem 'simple_form', github: 'plataformatec/simple_form', branch: 'v2.1'
```

I think that should work for you, thanks!

srpouyet:
@carlosantoniodasilva thanks, I searched in the issues but missed #1226. Cheers for the prompt reply!

Issue #1249, opened 2015-05-06
Missing doc for `selected` option for select inputs


PikachuEXE:
I tried to put `selected` option in `input_html` option
But actually it should be passed in directly
There is no doc mentioning available options and where to put them

rafaelfranca:
Yes. Could you submit a pull request?

PikachuEXE:
#1250 opened

feliperenan:
I'm closing this issue since it was resolved.

PR #1250, opened 2015-05-07
~ Add doc for option `selected` to README

PikachuEXE:
[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,8 +327,9 @@ translation. All other options given are sent straight to the underlying helper.
 can give prompt as:
 
 ```ruby
-f.input :age, collection: 18..60, prompt: "Select your age"
+f.input :age, collection: 18..60, prompt: "Select your age", selected: 21
 ```
+Extra options are passed into helper [`collection_select`](http://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-collection_select).
 
 It is also possible to create grouped collection selects, that will use the html *optgroup* tags, like this:
 


PikachuEXE:
This is for #1249

PikachuEXE:
@rafaelfranca PING!

README.md
```diff
@@ -346,6 +346,21 @@ used to retrieve label/value attributes for the `option` tags. Besides that, you
...
+# Note that the value passed in `selected` is an id, not an object
+f.input :parent_id, collection: Parent.all, selected: child.parent_id
+# "Multiple" select
+# Note that `:multiple` is passed inside `:input_html`
+f.input :children_ids, collection: Child.all, selected: parent.children_ids, input_html: {multiple: true}
+# Proc/Lambda
+# The value of `selected` is passed into rails helper without manupulation, so check its doc/code for detail
+f.input :caught_pokemon_type_ids, collection: Pokemon::Type.all, 
+  selected: lambda {|type| trainer.caught_pokemons.map(&:type_id)}
+```
```
carlosantoniodasilva: Thanks, but I think we don't need to show all possibilities here, since this is actually a functionality provided by Rails. I think it might make more sense to extend the prompt example on line 326, with something like this:

    All other options given are sent straight to the underlying helper. For example, you can give prompt as:

    ```ruby
    f.input :age, collection: 18..60, prompt: "Select your age"
    ```

    Or you can set the selected option as well:

    ```ruby
    f.input :age, collection: 18..60, selected: 21
    ```

    You can check the `collection_select` Rails documentation for more info on the available options.
```

And this last part that points to collection select rails docs could be a link to the api or something. Wdyt?

**Comment by PikachuEXE** [comment]
Done
Rails doc doesn't specify what can be accepted in `selected` though (But it's their fault)

File: README.md [hunk 2]
```diff
@@ -330,6 +330,14 @@ can give prompt as:
 f.input :age, collection: 18..60, prompt: "Select your age"
 ```
 
+Or you can set the selected option as well:
+
+```ruby
+f.input :age, collection: 18..60, selected: 21
+```
+The input value in option `select` will be passed into rails helper without modification.
```
carlosantoniodasilva: I think you meant `selected`? We don't need to make this paragraph specific to selected, as we can simply say that all these extra options are passed to collection select.

PikachuEXE:
Changed again

carlosantoniodasilva:
Thanks.

PR #1256, opened 2015-05-21
Try convert @object to model in case we got decorated object

timurvafin:
SimpleForm expects `@object` will be ActiveRecord instance and trying to call
- `@object.send(attribute_name)`
- `@object.column_for_attribute(attribute_name)`
- `@object.class.reflect_on_association(association)`

In case we use decorated (presenter) object when passing to `simple_form_for` association input does not work as expected.

Idea is to follow `@object.to_model` convention when trying to interact with `@object` as ActiveRecord instance, like Rails does in [FormHelper](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L458) for example.

I will add tests if idea generally is OK.

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -39,6 +39,7 @@ module SimpleForm
 
     def initialize(*) #:nodoc:
       super
+      @object   = convert_to_model(@object)
       @defaults = options[:defaults]
       @wrapper  = SimpleForm.wrapper(options[:wrapper] || SimpleForm.default_wrapper)
     end
--- a/test/form_builder/association_test.rb
+++ b/test/form_builder/association_test.rb
@@ -15,6 +15,12 @@ class AssociationTest < ActionView::TestCase
     end
   end
 
+  test 'builder association works with decorated object responsive to #to_model' do
+    assert_nothing_raised do
+      with_association_for @decorated_user, :company
+    end
+  end
+
   test 'builder association with a block calls simple_fields_for' do
     simple_form_for @user do |f|
       f.association :posts do |posts_form|
--- a/test/form_builder/error_test.rb
+++ b/test/form_builder/error_test.rb
@@ -36,6 +36,11 @@ class ErrorTest < ActionView::TestCase
     assert_select 'span.error', "cannot be blank"
   end
 
+  test 'error generates messages with decorated object responsive to #to_model' do
+    with_error_for @decorated_user, :name
+    assert_select 'span.error', "cannot be blank"
+  end
+
   test 'error generates messages for attribute with one error when using first' do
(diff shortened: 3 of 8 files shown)


carlosantoniodasilva:
I don't see any harm in doing that I think, please do add some tests for it. I wish Rails would pass the builder the converted object instead, since it's already doing the work to be able to figure out the necessary options- maybe that's another path to follow.

timurvafin:
@carloIsantoniodasilva I will add specs for the current implementation.

The bad thing with Rails that it converting object in several places in case it needs to work with object as an AR:B instance.
* https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L458
* https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L1909
* https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L1932
* https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/form_helper.rb#L1958

Don't know why they doing that way instead of converting in one place.

Thanks.

timurvafin:
@carlosantoniodasilva could you please take a look at the implementation with specs.

timurvafin:
@carlosantoniodasilva any thoughts?

carlosantoniodasilva:
@timurvafin sorry, I'll get back to you this week, thanks!

kevin-jj:
@object is not availble

timurvafin:
@carlosantoniodasilva any chance to get some feedback?

carlosramireziii:
@timurvafin @carlosantoniodasilva Just curious what the status of this issue is. Definitely would be a useful enhancement for me.

timurvafin:
+1

vladimirbazhanov:
this would be very useful functionality for me too so +1

feliperenan:
This PR still seems to be relevant. Could you rebase you branch with the master branch?

After that I we can merge this PR, @rafaelfranca any thoughts?

feliperenan:
@rafaelfranca I think that's ok merge it even it's been a long while. Can I merge it?

timurvafin:
@felipegimenesptec @rafaelfranca rebased!

andrey-skat:
> Idea is to follow @object.to_model convention when trying to interact with @object as ActiveRecord instance, like Rails does in FormHelper for example.

But they do it just for specific cases like building route by model and don't preserve result of conversion.
And now I can use `form_for` with my form object, but can't use simple form. 
If I make `def to_model; self; end` then I break automatic routing, if I don't do that, than I lost access to form object methods.

At least there should be access to original object. 
What do you think about revert this change or add `@original_object` method (or something)?

PR #1259, opened 2015-05-23
Use `button` instead of `submit` when a block is given.

lowjoel:
The submit helper does not accept blocks, and this unifies the behaviour of the different button types.

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -192,6 +192,23 @@ module SimpleForm
       input(attribute, options.merge(reflection: reflection))
     end
 
+    # Creates a submit button.
+    #
+    # This augments the original button implementation to generate a button element
+    # with a submit action when a block is given. Otherwise, it falls back to the
+    # original submit helper.
+    def submit_button(*args, &block)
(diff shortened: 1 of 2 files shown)


carlosantoniodasilva:
I feel it might be confusing that using submit would expect an input submit but will actually emit a button tag. I do like the idea of being able to pass a block, however you could already achieve that by using the `:button` option instead, right?

I'm considering we should keep users aware of the difference between button vs input, as long as there is a difference, so I'm giving this a close. Thank you for your contribution!

lowjoel:
If the user didn't give a block, this would still give an `<input>`. If I used `:button`, then the button would not have the characteristics of a submit button -- I will have to sprinkle all such buttons with a `type` and with a `name`. Not very DRY.

I think this would allow a clear separation between intent and presentation. What we are saying with the DSL is that we want a submit button. If a block is given, we want the button to have the given value. If a `<button>` is needed to preserve that intent, then a button should be emitted. The whole `<input>` vs `<button>` is nasty, I agree -- but I think the point of a DSL is to abstract away such nastiness.

PR #1260, opened 2015-05-23
Prefer the form-specified wrapper over the global mapping when it is specified

### LINKED ISSUE #1258: input wrapper decision order should prefer form-specific wrapper over mappings
Currently, the wrapper for a field is defined as:

 1. The specific type of wrapper specified on the input call
 2. The default wrapper mapping
 3. The current form's wrapper

It feels counterintuitive that the current form's wrapper takes lower precedence than the defaults configured in the initialiser.

If this is indeed believed to be a bug I should be able to file a PR to solve it.

lowjoel:
Fixes #1258.

I'm not sure where in the tests this should go, guidance welcome.

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -38,8 +38,9 @@ module SimpleForm
 
     def initialize(*) #:nodoc:
       super
-      @defaults = options[:defaults]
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request but we will stick with the current behavior. See https://github.com/plataformatec/simple_form/issues/1258#issuecomment-112382721

PR #1269, opened 2015-06-21
Open activemodel and actionpack dependencies to enable Rails5 support

hedgesky:
Current dependencies doesn't permit to use simple_form with edge Rails, so change them a bit.


hedgesky:
It's strange that Travis build failed: commands that failed there work perfectly in my console.

rafaelfranca:
It is already done in master.

rafaelfranca:
Thank you for the pull request.

PR #1279, opened 2015-07-06
Add example on how to set html attributes on specific collection items

obfuscoder:
When searching the internet I came upon "helpful" hints like that one:

http://stackoverflow.com/questions/17554748/how-do-i-add-html-attributes-to-select-options-with-simple-form-rails

I took quite some time before I figured that this was not the right solution. To not let others fall into this trap, I propose to add this to the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,7 +329,20 @@ can give prompt as:
 ```ruby
 f.input :age, collection: 18..60, prompt: "Select your age", selected: 21
 ```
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request. The readme already point to the Rails documentation that talks about this and more possibilities so I don't think duplicate the information here.

PR #1281, opened 2015-07-10
Remove nested (slow) i18n lookups

diegorv:
SimpleForm looks up for translation in a huge amount of places which
slows it down when you use nested forms.

See issue: plataformatec#1227

Removed the nested lookups code (including the while-loop) completely

### DIFF
--- a/lib/simple_form/inputs/base.rb
+++ b/lib/simple_form/inputs/base.rb
@@ -173,22 +173,10 @@ module SimpleForm
         model_names = lookup_model_names.dup
         lookups     = []
 
-        while !model_names.empty?
-          joined_model_names = model_names.join(".")
-          model_names.shift
+  lookups << :"#{model_names.join(".")}.#{reflection_or_attribute_name}"
+  lookups << default
 
-          lookups << :"#{joined_model_names}.#{lookup_action}.#{reflection_or_attribute_name}"
-          lookups << :"#{joined_model_names}.#{lookup_action}.#{reflection_or_attribute_name}_html"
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Hmm, I think the slowdown was caused by https://github.com/plataformatec/simple_form/pull/1024.

Thank you for investigating!

carlosantoniodasilva:
Yikes! Thanks @diegorv.

We should _at least_ make that optional and disabled by default, to avoid the perf hit for most people.

lib/simple_form/inputs/base.rb
```diff
@@ -173,22 +173,10 @@ def translate_from_namespace(namespace, default = '')
...
-        lookups << :"defaults.#{lookup_action}.#{reflection_or_attribute_name}"
-        lookups << :"defaults.#{lookup_action}.#{reflection_or_attribute_name}_html"
-        lookups << :"defaults.#{reflection_or_attribute_name}"
-        lookups << :"defaults.#{reflection_or_attribute_name}_html"
-        lookups << default
+  lookups << :"#{model_names.join(".")}.#{reflection_or_attribute_name}"
+  lookups << default
 
-        t(lookups.shift, scope: :"#{i18n_scope}.#{namespace}", default: lookups).presence
+  t(lookups.shift, scope: :"#{i18n_scope}.#{namespace}", default: lookups).presence
```
Yopi: These lines should be probably indented with the rest of the file.

rafaelfranca:
Fixed in 3.2.0

rafaelfranca:
Thank you for working on it

PR #1289, opened 2015-07-31
Allow no errors hash

sringling:
If the object used by the form doesn't have the typical ActiveModel::Errors collection, simple_form throws exceptions.  Applications might be using objects that are not from AR, like in cases that they are SOA objects.  If the application has decided to use some other method of showing errors, or is even just using a hash instead of a full-blown ActiveModel::Errors object, simple_form should allow it.

### DIFF
--- a/lib/simple_form/components/errors.rb
+++ b/lib/simple_form/components/errors.rb
@@ -38,19 +38,29 @@ module SimpleForm
       end
 
       def errors_on_attribute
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Thank you for the pull request. Simple Form is very tied to active record and right now we don't plan to change that. Although if you still plan to use simple form I think it will be better to you define your own error component. 

Thank you.

PR #1291, opened 2015-08-01
Use symbols instead of strings for #send

danielevans:
A tiny performance enhancement. Using symbol interpolation instead of strings is about 50% faster when being used with `#send`. The overall enhancement is very small since these methods are being called relatively rarely.

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -207,8 +207,8 @@ module SimpleForm
       options = args.extract_options!.dup
       options[:class] = [SimpleForm.button_class, options[:class]].compact
       args << options
-      if respond_to?("#{type}_button")
-        send("#{type}_button", *args, &block)
+      if respond_to?(:"#{type}_button")
+        send(:"#{type}_button", *args, &block)
       else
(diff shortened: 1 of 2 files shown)


carlosantoniodasilva:
I'm happy to merge it, but I'll have to add a side note that it's good to send benchmarks attached to code that is supposed to change performance in any way.

Thank you for your contribution! :)

danielevans:
Thanks! Apologies for the lack of benchmarks. I excluded them because running the gem's test suite showed no meaningful change. However, here are the benchmarks I used to convince me to make the change.

```ruby
require 'benchmark'

Benchmark.bm do |bm|
  bm.report('string') { 100_000_000.times { Object.send( "to_#{'s'}") } }
  bm.report('symbol') { 100_000_000.times { Object.send(:"to_#{'s'}") } }
end
```

```
       user     system      total        real
string 40.890000   0.070000  40.960000 ( 41.019328)
symbol 27.140000   0.030000  27.170000 ( 27.197318)
```

carlosantoniodasilva:
Thanks for sharing the benchmarks! :)

PR #1295, opened 2015-08-24
Yield options instead of options[:label] for label_text

dwilkie:
This gives more flexibility when overriding `label_text`

E.g.

``` ruby
config.label_text = lambda { |label, required, options| label_text =  "#{required} #{label}"; options[:icon] ? fa_icon(options[:icon], :text => label_text) : label_text }
```

The old `explicit_label` functionality can still be used

``` ruby
config.label_text = lambda { |label, required, options| explicit_label = options[:label].present? }
``
```

### DIFF
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -84,7 +84,7 @@ See https://github.com/plataformatec/simple_form/pull/997 for more information.
 
   # How the label text should be generated altogether with the required text.
   mattr_accessor :label_text
-  @@label_text = lambda { |label, required, explicit_label| "#{required} #{label}" }
+  @@label_text = lambda { |label, required, options| "#{required} #{label}" }
 
   # You can define the class to be used on all labels. Defaults to none.
   mattr_accessor :label_class
--- a/lib/simple_form/components/labels.rb
+++ b/lib/simple_form/components/labels.rb
@@ -33,7 +33,7 @@ module SimpleForm
 
       def label_text(wrapper_options = nil)
         label_text = options[:label_text] || SimpleForm.label_text
-        label_text.call(html_escape(raw_label_text), required_label_text, options[:label].present?).strip.html_safe
+        label_text.call(html_escape(raw_label_text), required_label_text, options).strip.html_safe
       end
 
       def label_target
--- a/test/form_builder/label_test.rb
+++ b/test/form_builder/label_test.rb
@@ -107,15 +107,22 @@ class LabelTest < ActionView::TestCase
     end
   end
 
+  test 'label_text yields with options' do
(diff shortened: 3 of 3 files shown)


rafaelfranca:
Thank you so much for the pull request. Unfortunately this is a breaking change and I don't see a compiling reason to accept it. It is already possible to do what you want using a custom label component and I think this is the way to go.

dwilkie:
Thanks for the feedback. I was trying to find a way to what I need to do using a custom component as you suggested but I couldn't find a clean way to do it.

I want to use all of labels I18n lookup functionality, but add an icon with the translation text as the label. This worked fine before when I had the html for the label and icon directly in the translation yml files. However simple_form now escapes this html.

I need the following HTML:

```html
<label class="email required control-label" for="user_email"><abbr title="ទាមទារ">*</abbr> <i class="fa fa-envelope"></i> អ៊ីម៉ែល</label>
```

Basically I just need to replace the translated email text with: `<i class="fa fa-envelope"></i> អ៊ីម៉ែល`

This would work perfectly if I could set `config.label_text = lambda { |label, required, options| label_text =  "#{required} #{label}"; options[:icon] ? fa_icon(options[:icon], :text => label_text) : label_text }`.

Failing that how can I cleanly use a custom label component with all the same functionality as the built in label component?

rafaelfranca:
You can either open the label component and change the `label_text` method as your wish or you can create a new component that includes the actual label component and inherit all the behavior.

dwilkie:
Ok so I have to monkey patch `label_text` or make a new component and include the actual label component. If I make a new label component then I have to create a new wrapper that contains exactly the same content as the label wrapper and replace `ba.use :label` with `ba.use :custom_label`. This just seems a bit overkill when we have the `label_text` lambda.

I like the idea of the `label_text` lambda which can be overridden for this exact purpose. It would just be better if we could yield to it with the options instead of `options[:label].present?`. Anyway I understand that this is a potentially breaking change if people have overridden this block in their applications.

Would you accept the pull request if I add `options` as a 4th argument and check the arity of the proc before calling it?

rafaelfranca:
In fact I regret for not having options as the third argument since the beginning. But using a wrapper based config like a custom component seems better to me that using a global config like `label_text`.

If we can think in a good deprecation strategy I'm more willing to merge the PR changing the third argument to be options @carlosantoniodasilva WDYT?

dwilkie:
For a deprecation strategy, we could call the proc with the options as the third arguement, catch any exception, show a deprecation warning and then call it again with the old `options[:label].present?`. Thoughts?

feliperenan:
I'm not sure with should go ahead with this PR since it hasn't recent activity. Can I close this one?

cc @rafaelfranca

feliperenan:
Well, I'm gonna this PR. Case this still an issue and someone wants to work on it, feel free to reopen.

PR #1301, opened 2015-08-31
Merge attributes with wrapper_options

### LINKED ISSUE #1278: merge_wrapper_options - missing keys
Maybe it's just a misunderstanding but I've figured out something wrong with `merge_wrapper_options` based on this pull request: https://github.com/hughevans/pagedown-bootstrap-rails/pull/11 As you can see we lost the keys after using the method. Shouldn't we consider this case?

Comments are welcome.

herminiotorres:
As described in the #1278 issue modified the method of `merge_wrapper_options` to support duplicated values.

- If sending a new wrapper it returns the new.
- If sending a new wrapper with same attributes merge with the older.

closes #1278

### DIFF
--- a/lib/simple_form/inputs/base.rb
+++ b/lib/simple_form/inputs/base.rb
@@ -189,9 +189,14 @@ module SimpleForm
 
       def merge_wrapper_options(options, wrapper_options)
         if wrapper_options
-          options.merge(wrapper_options) do |_, oldval, newval|
-            if Array === oldval
-              oldval + Array(newval)
+          wrapper_options.merge(options) do |key, oldval, newval|
+            case key.to_s
+            when "class"
+              Array(oldval) + Array(newval)
+            when "data", "aria"
+              oldval.merge(newval)
+            else
+              newval
             end
           end
         else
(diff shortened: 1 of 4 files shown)


rafaelfranca:
:shipit:. Please add a CHANGELOG entry.

rafaelfranca:
We need to make the test conditional to the Rails version.

```ruby
if ActionPack::VERSION::MAJOR == '4' && ActionPack::VERSION::MINOR >= '2'
```

rafaelfranca:
@herminiotorres :shipit:

lib/simple_form/inputs/base.rb
```diff
@@ -189,9 +189,14 @@ def translate_from_namespace(namespace, default = '')
 
       def merge_wrapper_options(options, wrapper_options)
         if wrapper_options
-          options.merge(wrapper_options) do |_, oldval, newval|
-            if Array === oldval
-              oldval + Array(newval)
+          wrapper_options.merge(options) do |key, oldval, newval|
+            case key.to_s
+            when "class"
```
lucasmazza: @rafaelfranca makes sense to freeze these Strings? :snowflake:

rafaelfranca: I prefer to not, unless we have proof that this will improve significantly the performance of the application.

PR #1312, opened 2015-10-17
Fix input_field not receiving the wrapper options

### LINKED ISSUE #1230: SimpleForm::FormBuilder#Input_field produces a non styled input when using with twitter bootstrap
Example code:

``` haml
= f.input :start_at, :label => "Start/End dates" do
  = f.input_field :start_at, :as => :date_picker
  = f.input_field :end_at, :as => :date_picker
```

Results in:
simple_form

To achieve the expected result I had to do:

``` haml
= f.input :start_at, :label => "Start/End dates" do
  = f.input_field :start_at, :as => :date_picker, :class => "form-control pull-left"
  = f.input_field :end_at, :as => :date_picker, :class => "form-control pull-left"
```

Which doesn't seem right to me since I have to repeat bootstrap-specific class name each time.

Seems like a bug to me.

amoludage:
Previously any option given to the wrapper `:input` config, like a class
name or data attribute, would not be passed down when generating an
input field. Now if you give any specific option to generate inputs with,
they will be also be picked up when using `input_field`.

Fixes #1230.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### bug fix
   * `date/time/datetime` inputs now correctly generate the label `for` attribute when
   HTML5 compatibility is explicitly enabled. [@ericsullivan](https://github.com/ericsullivan)
+  * `input_field` now correctly generates extra options from the wrapper input config.
 
 ## 3.1.0
 
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -137,9 +137,15 @@ module SimpleForm
       options[:input_html] = options.except(:as, :boolean_style, :collection, :label_method, :value_method, *components)
       options = @defaults.deep_dup.deep_merge(options) if @defaults
 
-      input      = find_input(attribute_name, options)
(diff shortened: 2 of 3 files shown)


cireficc:
@amoludage I'm confused by this fix. You have a [branch](https://github.com/plataformatec/simple_form/tree/ca-fix-input_field-wrapper)/[commit](https://github.com/plataformatec/simple_form/commit/8842c2f23015c62ed44d8c746a30376504f3a336) that fixes the issue but it's not merged into master? This was fixed last October, are there any plans for merging this?

A simple work-around is available, by passing any options you need for an `input_field`, but having a lot of `input_field`s that need the same options results in a **lot** of duplication in forms, whereas a simple wrapper could be used.

For example, duplicating:

```ruby
= f.input_field :field_name,
    as: :string,
    class: "form-control",
    ....
```

with several more attributes is form bloat, and it's not fun to maintain. I appreciate all the work put into the gem though, it's very useful!

l0c0luke:
ya +1

cerdiogenes:
+1

diegous:
+1

PR #1323, opened 2015-11-16
Remove unnecessary code

asiniy:
Make code style same as code below


carlosantoniodasilva:
Thanks! Unfortunately the `||=` check changes the logic slightly so that the `false` wrapper option passed to `simple_fields_for` would now default to the parent's wrapper option, so we have to explicitly check for `nil` instead. [This test represents this use case](https://github.com/plataformatec/simple_form/blob/71054e5623c53bb7d1323c27cb4271fb6e4f7da7/test/action_view_extensions/builder_test.rb#L606-L614).

PR #1327, opened 2015-11-28
Add item_extra_html option for collection check/radio buttons

nathansamson:
Rationale behind this change is to be able to easily add "data-" attributes to checkboxes/radiobuttons depending on the value.

We use this to dynamically (with javascript) enable / disable options in a list depending on a choice made in another list

If there are easy ways to achieve this with existing simple_form methods, I am happy to hear them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.X.X
+
+### enhancements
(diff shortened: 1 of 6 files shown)


feliperenan:
I'm closing this PR since it hasn't recently activity. We can reopen it for discussion case someone think it's a useful feature.

PR #1333, opened 2015-12-23
Add :valid_class on input wrapper when value is present and valid

aeberlin:
##### Purpose

The purpose of this PR is to improve style support for inputs with validations. Simply put, the lack of an applied `:error_class` is not indicative of the attribute being valid. While this isn't a guaranteed method of determining if it is valid, it is as close as programmatically possible (AFAIK). See caveats below.
##### Logic

When a `:valid_class` is specified, it will be applied to an input wrapper given the following conditions:
1. Object exists.
2. Object responds to `:errors`.
3. There are no validation errors (e.g. validations passed or were not specified).
4. Object responds to the attribute method.
5. Return value of attribute method is present.
##### Changes
1. Added `:valid_class` declarations (similar to those for `:error_class`) on default wrappers.
2. Added `:has_value?` and `:valid?` to `SimpleForm::Components::Errors`.
3. Adjusted `SimpleForm::Wrappers::Root#html_classes` to apply `:valid_class` when applicable.
##### Known caveats

If an attribute is valid when `:blank?`, the `:valid_class` will not be applied. I am open to suggestions here, as this is the only hole in my logic, albeit an acceptable and predictable one (IMO).
##### Further notes

I adjusted two failing tests that were missing stubs, and added another test for this feature.

Please let me know if there's something else I should include. Thanks, cheers!

### DIFF
--- a/lib/simple_form/components/errors.rb
+++ b/lib/simple_form/components/errors.rb
@@ -13,6 +13,16 @@ module SimpleForm
         object && object.respond_to?(:errors) && errors.present?
       end
 
+      def has_value?
+        return unless object && object.respond_to?(attribute_name)
+
+        object.send(attribute_name).present?
+      end
+
+      def valid?
+        !has_errors? && has_value?
+      end
+
       protected
 
       def error_text
--- a/test/form_builder/error_test.rb
+++ b/test/form_builder/error_test.rb
@@ -243,4 +243,16 @@ class ErrorTest < ActionView::TestCase
       assert_no_select 'span.error'
     end
   end
+
+  # NO ERRORS
+
+  test 'input wrapper has valid class for present attribute without errors' do
+    @user.instance_eval do
+      undef errors
+      name = 'foo'
+    end
+
+    with_form_for @user, :name
+    assert_select '.input.field_without_errors'
+  end
 end
--- a/README.md
+++ b/README.md
@@ -838,7 +838,8 @@ The syntax looks like this:
 
 ```ruby
 config.wrappers tag: :div, class: :input,
-                error_class: :field_with_errors do |b|
+                error_class: :field_with_errors,
+                valid_class: :field_without_errors do |b|
 
   # Form extensions
   b.use :html5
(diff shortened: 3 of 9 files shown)


aeberlin:
@plataformatec Feedback, please?

test/form_builder/error_test.rb
```diff
@@ -243,4 +243,16 @@ def with_full_error_for(object, *args)
       assert_no_select 'span.error'
     end
   end
+
+  # NO ERRORS
```
feliperenan: I think you can remove this comentary

lib/simple_form/components/errors.rb
```diff
@@ -13,6 +13,16 @@ def has_errors?
         object && object.respond_to?(:errors) && errors.present?
       end
 
+      def has_value?
+        return unless object && object.respond_to?(attribute_name)
```
feliperenan: As we expect a boolean here, we can return `false`, What do you think?

m5o:
@feliperenan would you may take over this PR? It seems @aeberlin is not very active, except probably no response. Did he allow PR edits from maintainers? Otherwise, I would volunteer, with your guidance.

rafaelfranca:
👍 I'm good with the idea

feliperenan:
@m5o We can't commit on this branch as well. If you are willing in working on it go ahead, just open another pull request then we can close this one.

lib/simple_form/components/errors.rb
josevalim: The whole method could be a a single operation:

    object && object.respond_to?(attribute_name) && object.send(attribute_name).present?

m5o: @josevalim - Done ✅  https://github.com/plataformatec/simple_form/pull/1553/commits/12f38eeaa0ea392a7d3ba75796928437660a3577

feliperenan:
closing in favor https://github.com/plataformatec/simple_form/pull/1553

PR #1335, opened 2015-12-23
Updated README.md to reflect that gender is not binary

lblackburn:
https://gist.github.com/garbados/f82604ea639e0e47bf44


rafaelfranca:
I prefer to avoid this kind of discussion in the future and change entirely that example. Could you change `:gender` for something else?

PR #1342, opened 2016-01-10
fix decimal_or_float? method (undefined method `number?` in rails 5)

kesha-antonov:


lucasmazza:
@rafaelfranca do you want to double check this one? I'm not sure if we can write a reliable test case right now as our test suite doesn't hit any db.

rafaelfranca:
For Rails 5 we should stop relying in column and use the type information. With that we don't need to change any code since `number?` is now in the type object.

olivierlacan:
@rafaelfranca How does that type object get accessed?

rafaelfranca:
@olivierlacan [`type_for_attribute`](http://edgeapi.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-type_for_attribute)

pokonski:
Is there any official fix for it yet? Rails 5 is already in RC1 but the issue is still present.

olivierlacan:
This PR appears abandoned, I'm going to submit a new one if you folks don't mind.

pokonski:
That'd be great, since there's already Rails 5 stable.

lucasmazza:
@oliverlacan feel free to do it. I think we can use `type_for_attribute` inside `FormBuilder#find_attribute_column` to start using the Attributes API but I don't know how backwards compatible it is with the existing column types we currently use.

olivierlacan:
@lucasmazza There's already a method check in there so I can just ask the receiver if it has `type_for_attribute` and go from there to try to keep it backward-friendly.

pcantrell:
Given that this is a showstopper bug for Rails 5, may I put in a vote for merging this very old PR now, and working out whether to use `type_for_attribute` in a separate PR?

lucasmazza:
I've merged #1417 and release Simple Form `3.3.0` with it and the extra patches we had pending on master - thanks @olivierlacan for the initial patch. Let us know if Simple Form doesn't cover any usage case for Rails 5 so we can patch it up.

olivierlacan:
Sorry for the lack of follow-up on this @lucasmazza and thanks for taking care of this. 👍

PR #1345, opened 2016-01-12
Browsers do not support \A or \Z, but rails recommends them

singpolyma:
So translate to something close that the browser supports


lucasmazza:
@singpolyma can you please add a test for this change? :)

singpolyma:
@lucasmazza test added

feliperenan:
I'm not sure if we should merge this PR since it hasn't recent activity and it seems like not a common use case. So, I'm gonna close this PR by now.

Thanks

PR #1368, opened 2016-03-08
Add a way to specify custom collection option lookup key for i18n

doits:
A custom key can specified with `config.i18n_options_scope`:

    config.i18n_options_scope = lambda { |i18n_scope| "#{i18n_scope}.options" }

The default (above) returns `simple_form.options` - like it was always
before. But I can no easy override the `options` key like this:

    config.i18n_options_scope = lambda { |i18n_scope| "#{i18n_scope}.other" }

In addition, if I want it to be somewhere else outside the regular
`i18n_scope`, I can do:

    config.i18n_options_scope = lambda { |_| 'where.ever.i.want' }

This gives total flexibility where simple_form looks up translations of
collection options.

The current version always wants the config to be a `lambda`. One could
rewrite the implementation to work with a simple string
(`config.i18n_options_scope = 'where.ever.i.want`) if we want to give
the user total ease of defining the scope in case the scope does not
depend on the `i18n_scope`. Or the implementation could be that in case
of a single string the `i18n_scope` is automatically prepended, for
example:

    config.i18n_options_scope = 'where.ever.i.want'

... means to look up under `simple_form.where.ever.i.want`. But for
explicitness sake requiring it to be a `lambda` always might be the
better choice (but I'm open to changes).

### DIFF
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -202,6 +202,10 @@ See https://github.com/plataformatec/simple_form/pull/997 for more information.
   mattr_accessor :i18n_scope
   @@i18n_scope = 'simple_form'
 
+  # Where to look for translation of collection options
+  mattr_accessor :i18n_options_scope
+  @@i18n_options_scope = lambda { |i18n_scope| "#{i18n_scope}.options" }
+
   # Retrieves a given wrapper
   def self.wrapper(name)
     @@wrappers[name.to_s] or raise WrapperNotFound, "Couldn't find wrapper with name #{name}"
--- a/lib/simple_form/inputs/base.rb
+++ b/lib/simple_form/inputs/base.rb
@@ -169,7 +169,7 @@ module SimpleForm
       #            email: 'E-mail.'
       #
       #  Take a look at our locale example file.
-      def translate_from_namespace(namespace, default = '')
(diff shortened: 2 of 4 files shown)


doits:
@plataformatec - any chance for a review/comment?

lucasmazza:
@doits sorry for the delay - what's the motivation behind this new configuration? Isn't the default i18n scope enough for your application? Would love to know more about to see if we should merge this or if the default behavior should be enough.

doits:
It is simple: I am using some other libraries that work with and lookup translations of collection values. So I already have defined them at a different namespace (in my case called `activerecord.attribute_values` - library name agnostic) and do not want to define them again but make simple form use them.

I could work round this if there was a way to set translations during runtime, something like:

```
I18n.available_locales.each do |loc|
  I18n.store_translation("#{loc}.simple_form.options", I18n.t("#{loc}.activerecord.attribute_values"))
end
```

This could be in an initializer and all was good. Do you maybe know a way to do something like this? Then this PR would be superfluous.

doits:
I think I found a solution to do it with `I18n.backend.store_translations`. So this can be closed unless `store_translations` does not work - will reopen it then.

PR #1369, opened 2016-03-10
Spaces between labels and checkboxes for Twitter Bootstrap 4

woto:


rafaelfranca:
Thank you for the pull request. We avoid to change Simple Form code to add support to any framework. Is not possible to add this same space with the wrapper API or even with CSS?

woto:
Hi, i don't know for now. I foresaw this answer :), just pointed to this issue. May be somebody will find better solution.

xijo:
@rafaelfranca What about joining with `"\n"` instead? It solves the problem, but does not introduce unnecessary "visible" characters.

m5o:
additional styling could be a workaround:

```scss
// vertical form
.form-check {
  > .collection_check_boxes,
  > .collection_radio_buttons {
    @extend .form-check-label;
    // fix missing space between input & label
    margin-right: .265625rem;
  }
}
```

rafaelfranca:
I think it is fine to add the space by default. Can someone add tests and update this branch?

m5o:
Yes, it would be useful. 
Currently this is patched by CSS adding margin to the `check_boxes` / `radio_buttons`

🔗  [`_form_collection_label.scss#L6-L7`](https://github.com/rafaelfranca/simple_form-bootstrap/blob/master/app/assets/stylesheets/simple_form-bootstrap/_form_collection_label.scss#L6-L7)
:link: [`_form_collection_label.scss#L16-L17`](https://github.com/rafaelfranca/simple_form-bootstrap/blob/master/app/assets/stylesheets/simple_form-bootstrap/_form_collection_label.scss#L16-L17)

feliperenan:
I tried to reproduce this issue on https://github.com/rafaelfranca/simple_form-bootstrap but I couldn't. I didn't see the difference and, pointing that project to use this branch the issue is still there if I remove the styles mentioned by @m5o. Could someone help me with that?

Also, I realized that this patch only fixes the boolean inputs, checkbox and radio buttons that are `:nested`. Should it be applied to inputs that are `:inline` as well?

m5o:
I've tried to point this branch to latest `simple_form-bootstrap` without success. It seems so outdated with current `simple_form` master. So I've patched the changes into my local `gems/simple_form-4.1.0/` source and can confirm @feliperenan findings, the `" "` or any other string I've tested is exposed to the view. So I've reverted the changes.

#### Status quo

The following gif shows the view toggle between bootstrap and simple_from markup in Chrome v72. The css fix (`margin-right: .265625rem;`) is **not** applied.
simple_form-inline

If, for example, you select the bootstrap markup you could literally see the `space character` between the inline check boxes. ☝️The space is after, not between.

screenshot 2019-01-30 at 12 52 02

#### Progress

If there is a patch needed, then how the inline radio and checkboxes are concatenated. And I think the solution is not a space between the checkbox and the label. 

If you take a look at the raw source code of `simple_form` inline radio collection. The whole `fieldset` html source is concatenated without a space or newline between. If you add a newline or space character before the `<div class="form-check form-check-inline">` element it will become visually equal to the bootstrap output.

```html
<!-- space -->
[…]<label class="collection_check_boxes" for="user_fruit_apple">Apple</label></div> <div class="form-check form-check-inline">[…]
```

###### newline
screenshot 2019-01-30 at 13 17 52

I don't know if I'm right, hope this feedback helps somehow. ✌️ 

🚀 With or without a space, I guess this will not bring `simple_form` forward. If you look at the other css workarounds that are needed to radio and checkboxes it would be more important to allow customize the currently fixed css classes by simple_form. eg. (`.radio_buttons` / `.check_boxes` / `.collection_radio_buttons` / `.collection_check_boxes`)

feliperenan:
@m5o According to the gif that you've shown, I'm not sure if we are facing the same problem.

* Form without this patch [`_form_collection_label.scss#L6-L7`](https://github.com/rafaelfranca/simple_form-bootstrap/blob/master/app/assets/stylesheets/simple_form-bootstrap/_form_collection_label.scss#L6-L7)

image


Even adding an empty string between the label and radio button, we'll face the same issue.

I noticed that we are not using the class `form-check-label`, and I think that's the issue. Digging into Simple Form code, I realized that we have `item_label_class` option, but we use this attribute only when checkboxes and/or radio buttons are `:nested`. We never use it when they are `:inline`.

https://github.com/plataformatec/simple_form/blob/7cbedde3de8501946fb85e5e8a7ecb767fb5c0ff/lib/simple_form/tags.rb#L19

Issue related: https://github.com/plataformatec/simple_form/issues/1628#issuecomment-452304605

I think that with we be able to set `item_label_class` for :inline checkboxes / radio buttons as well we solve this issue. Does it make sense?

m5o:
You probably also removed line 5 of [`_form_collection_label.scss#L5`](https://github.com/rafaelfranca/simple_form-bootstrap/blob/master/app/assets/stylesheets/simple_form-bootstrap/_form_collection_label.scss#L5), the `@extend .form-check-label;` is fixing the vertical miss alignment of the radio. 😉

The `item_label_class` option for inline collections is definitely useful and will help reduce css workarounds. 👍

feliperenan:
> You probably also removed line 5 of _form_collection_label.scss#L5, the @extend .form-check-label; is fixing the vertical miss alignment of the radio. 

Ohh, you are right 🙈.

So, I'm gonna close this PR. Thanks for the help 🎉

PR #1374, opened 2016-04-01
Add a horizontal-multi-select wrapper for Bootstrap forms

rtrv:
Default multi-select is for vertical forms. New wrapper allows using date and time multi-selects in horizontal forms.


feliperenan:
It was added with [Bootstrap's update](https://github.com/plataformatec/simple_form/pull/1569)

PR #1376, opened 2016-04-15
Add `hidden_input` option to checkbox

capripot:
Give the possibility to skip the creation of the hidden input field
normally created along with checkox field, replicating [Rails 4.2 behavior](https://github.com/rails/rails/blob/4-2-stable/actionview/lib/action_view/helpers/tags/check_box.rb#L28)


oswaldoferreira:
Any maintainer to take a look?

lucasmazza:
Looks good to me, but @rafaelfranca could do a second check just in case.

(If anyone is curious, this was added on rails back in rails/rails#5420)

Issue #1382, opened 2016-05-13
allow_blank: true validator not handled


seliverstov-maxim:
SimpleForm builds an 'required' input with for the "presence:true, allow_blank: true" validatior.

For example:

class User < ActiveRecord::Base
validates :password, presence: true, allow_blank: true
end

in view
= f.input :password

html
`<div class="form-group password required user_password"><label class="password required control-label" for="user_password"><abbr title="required">*</abbr> Password</label><input class="password required form-control" type="password" name="user[password]" id="user_password" /></div>`

seliverstov-maxim:
Fixed in: https://github.com/plataformatec/simple_form/pull/1383

seliverstov-maxim:
No reason to fix it. Sorry for mistake.

PR #1383, opened 2016-05-13
Fix required input for allow_blank attribute

seliverstov-maxim:
See https://github.com/plataformatec/simple_form/issues/1382


seliverstov-maxim:
No reason to fix it. Sorry for mistake.

PR #1392, opened 2016-06-03
Adds aria-invalid attribute to fields with errors

amoose:
Adds `aria-invalid` attribute to fields with errors. 

https://www.w3.org/TR/WCAG20-TECHS/ARIA21.html


lucasmazza:
@amoose thanks :)

PR #1393, opened 2016-06-06
Add label_method info in README

flanger001:
From @rafaelfranca's StackOverflow answer here: http://stackoverflow.com/questions/6334582/simple-form-association-custom-label-name/6335679#6335679

This seems like fairly useful information and the comment below Rafael's seemed to indicate others would find it helpful in the docs also.


lucasmazza:
Thanks!

PR #1396, opened 2016-07-01
Add support Rails 5

herminiotorres:
Enhancements
  * Support to Rails `5.0.x` Stable.

Unreleased
  * Removed deprecated methods `default_input_size`.

### DIFF
--- a/lib/simple_form/components/maxlength.rb
+++ b/lib/simple_form/components/maxlength.rb
@@ -15,10 +15,7 @@ module SimpleForm
           maxlength
         else
           length_validator = find_length_validator
-
-          if length_validator && !has_tokenizer?(length_validator)
-            length_validator.options[:is] || length_validator.options[:maximum]
-          end
+          maximum_length_value_from(length_validator)
         end
       end
 
@@ -28,6 +25,22 @@ module SimpleForm
 
       def has_tokenizer?(length_validator)
         length_validator.options[:tokenizer]
+      end
+
+      # Use validation with tokenizer if version of Rails is less than 5,
+      # if not validate without the tokenizer, if version is greater than Rails 4.
+      if ActionPack::VERSION::STRING < '5'
+        def maximum_length_value_from(length_validator)
+          if length_validator && !has_tokenizer?(length_validator)
+            length_validator.options[:is] || length_validator.options[:maximum]
+          end
+        end
+      elsif ActionPack::VERSION::STRING >= '5'
+        def maximum_length_value_from(length_validator)
+          if length_validator
+            length_validator.options[:is] || length_validator.options[:maximum]
+          end
+        end
       end
     end
   end
--- a/test/inputs/discovery_test.rb
+++ b/test/inputs/discovery_test.rb
@@ -14,6 +14,7 @@ class DiscoveryTest < ActionView::TestCase
         Object.send :remove_const, :CustomizedInput
         Object.send :remove_const, :DeprecatedInput
         Object.send :remove_const, :CollectionSelectInput
+        CustomInputs.send :remove_const, :CustomizedInput
         CustomInputs.send :remove_const, :PasswordInput
         CustomInputs.send :remove_const, :NumericInput
       end
--- a/test/inputs/priority_input_test.rb
+++ b/test/inputs/priority_input_test.rb
@@ -5,14 +5,22 @@ class PriorityInputTest < ActionView::TestCase
   test 'input generates a country select field' do
     with_input_for @user, :country, :country
     assert_select 'select#user_country'
-    assert_select 'select option[value=Brazil]', 'Brazil'
+    if ActionPack::VERSION::STRING >= '5'
+      assert_select 'select option[value=BR]', 'Brazil'
+    elsif ActionPack::VERSION::STRING < '5'
+      assert_select 'select option[value=Brazil]', 'Brazil'
+    end
     assert_no_select 'select option[value=""][disabled=disabled]'
   end
 
   test 'input generates a country select with SimpleForm default' do
     swap SimpleForm, country_priority: [ 'Brazil' ] do
       with_input_for @user, :country, :country
-      assert_select 'select option[value=""][disabled=disabled]'
+      if ActionPack::VERSION::STRING >= '5'
+        assert_select 'select option[value="---------------"][disabled=disabled]'
+      elsif ActionPack::VERSION::STRING < '5'
+        assert_select 'select option[value=""][disabled=disabled]'
+      end
     end
   end
 
--- a/test/support/models.rb
+++ b/test/support/models.rb
@@ -215,7 +215,9 @@ class ValidatingUser < User
     only_integer: true
   validates_length_of :name, maximum: 25
   validates_length_of :description, maximum: 50
-  validates_length_of :action, maximum: 10, tokenizer: lambda { |str| str.scan(/\w+/) }
+  if ActionPack::VERSION::STRING < '5'
+    validates_length_of :action, maximum: 10, tokenizer: lambda { |str| str.scan(/\w+/) }
+  end
   validates_length_of :home_picture, is: 12
 
   def min_amount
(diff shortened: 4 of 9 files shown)


lucasmazza:
@herminiotorres looks like we have some conflicts from bd7127905009aa0c8028c9726b61aa38b031f20f here 🙈

herminiotorres:
@lucasmazza my branch was outdated, I just solved the conflict! :speak_no_evil:

gemfiles/Gemfile.rails-5-0-stable
```diff
@@ -0,0 +1,16 @@
...
+
+gemspec :path => '..'
+
+platforms :rbx do
+  gem 'rubysl', '~> 2.0'
+  gem 'rubinius-developer_tools'
+end
+
+gem 'country_select', '~> 2.5.2'
+gem 'railties', '5.0.0'
```
lucasmazza: The default `Gemfile` is already targeting the 5.0.0 releases, so we don't need a new Gemfile for it.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 * Updated gem dependency to support Rails 5.0.x.
 
+### enhancements
+  * Support to Rails 5.0.x Stable. [@herminiotorres](https://github.com/herminiotorres)
+  * Removed deprecated methods `default_input_size`. [@herminiotorres](https://github.com/herminiotorres)
```
lucasmazza: These changelog entries should be moved to the `## Unreleased` block.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
...
+  * Update `country_select` to version `2.5.2`. [@herminiotorres](https://github.com/herminiotorres)
```
lucasmazza: `country_select` isn't a runtime dependency here, so the value for reporting this back to developers that come to our CHANGELOG for updates on Simple Form isn't much helpful - the gem already works with `country_select`, we are just updating our test suite for it, right?

test/inputs/discovery_test.rb
```diff
@@ -14,6 +14,7 @@ def discovery(value = false)
         Object.send :remove_const, :CustomizedInput
         Object.send :remove_const, :DeprecatedInput
         Object.send :remove_const, :CollectionSelectInput
+        CustomInputs.send :remove_const, :CustomizedInput
```
lucasmazza: ❤ 💚 💙 💛 💜

test/inputs/priority_input_test.rb
```diff
@@ -5,14 +5,14 @@ class PriorityInputTest < ActionView::TestCase
...
-    assert_select 'select option[value=Brazil]', 'Brazil'
+    assert_select 'select option[value=BR]', 'Brazil'
     assert_no_select 'select option[value=""][disabled=disabled]'
   end
 
   test 'input generates a country select with SimpleForm default' do
     swap SimpleForm, country_priority: [ 'Brazil' ] do
       with_input_for @user, :country, :country
-      assert_select 'select option[value=""][disabled=disabled]'
+      assert_select 'select option[value="---------------"][disabled=disabled]'
```
olivierlacan: Is this necessary for Rails 5 support?

lucasmazza: I think it doesn't, but sounds like the `priority_divider` default from `country_select` changed between 1.x and 2.x.

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 ## Unreleased
 
 * Add the `aria-invalid` attribute on inputs with errors.
+* Removed deprecated methods `default_input_size`. [@herminiotorres](https://github.com/herminiotorres)
 
 ## 3.2.1
 
 * Updated gem dependency to support Rails 5.0.x.
 
+### enhancements
```
lucasmazza: We already have an entry on the 3.2.1 changelog for the Rails 5 dependency, so we can ✂️ this one out.

olivierlacan: Problem is this does not bring proper Rails 5 support due to `number?` issues in `decimal_or_float`. See #1342

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 ## Unreleased
 
 * Add the `aria-invalid` attribute on inputs with errors.
+* Removed deprecated methods `default_input_size`. [@herminiotorres](https://github.com/herminiotorres)
```
rafaelfranca: This doesn't need a changelog entry since the methods are still there, just not duplicated.

rafaelfranca:
I'll merge and cleanup the changelog later.

test/support/models.rb
```diff
@@ -215,7 +215,7 @@ class ValidatingUser < User
     only_integer: true
   validates_length_of :name, maximum: 25
   validates_length_of :description, maximum: 50
-  validates_length_of :action, maximum: 10, tokenizer: lambda { |str| str.scan(/\w+/) }
+  validates_length_of :words_in_action, maximum: 10
```
rafaelfranca: Hmmm, I think this may invalidate the test that is checking this. I think we can just remove it. Could you take a look?

herminiotorres: The test pass in all rails of versions. When I remove this code.

rafaelfranca: I mean, this test now doesn't make any sense:

https://github.com/plataformatec/simple_form/blob/81b9d7d3439225c7613e570afb44596acfa97a76/test/inputs/string_input_test.rb#L35-L38

We should remove the test when running with Rails 5 but keep for the others versions. The same with this code, we should only define this validation when it is not Rails 5. We should also conditionally define https://github.com/plataformatec/simple_form/blob/81b9d7d3439225c7613e570afb44596acfa97a76/lib/simple_form/components/maxlength.rb#L29-L31.

herminiotorres: :ok_hand:

herminiotorres: @rafaelfranca 
I applied your suggestion but still it displays this `deprecation` information. But see how it looks, that's what you suggested right?

```ruby
DEPRECATION WARNING: The `:tokenizer` option is deprecated, and will be removed in Rails 5.1.
You can achieve the same functionality by defining an instance method
with the value that you want to validate the length of. For example,

    validates_length_of :essay, minimum: 100,
      tokenizer: ->(str) { str.scan(/w+/) }

should be written as

    validates_length_of :words_in_essay, minimum: 100

    private

    def words_in_essay
      essay.scan(/w+/)
    end
 (called from <class:ValidatingUser> at /Users/herminiotorres/src/simple_form/test/support/models.rb:218)
```

rafaelfranca: In Rails 5 you don't need to define that validation, so just remove it for Raisl 5

lib/simple_form/components/maxlength.rb
```diff
@@ -29,6 +26,20 @@ def find_length_validator
       def has_tokenizer?(length_validator)
         length_validator.options[:tokenizer]
       end
+
+      # Use validation with tokenizer if version of Rails is less than 5,
+      # if not validate without the tokenizer, if version is greater than Rails 4.
+      def validation_with_tokenizer(length_validator)
+        if ActionPack::VERSION::STRING < '5'
```
rafaelfranca: Put the conditional outside the method to avoid runtime conditionals when we know the value in load time.

lib/simple_form/components/maxlength.rb
```diff
@@ -29,6 +26,20 @@ def find_length_validator
       def has_tokenizer?(length_validator)
         length_validator.options[:tokenizer]
       end
+
+      # Use validation with tokenizer if version of Rails is less than 5,
+      # if not validate without the tokenizer, if version is greater than Rails 4.
+      def validation_with_tokenizer(length_validator)
```
rafaelfranca: This name could be: `maximum_length_value_from`

herminiotorres:
:shipit: ? 😄

herminiotorres:
@lucasmazza :shipit: ?

rafaelfranca:
It is still missing the conditional in the validatation code here https://github.com/plataformatec/simple_form/blob/master/test/support/models.rb#L218

PR #1397, opened 2016-07-01
Make sure that builder is set.

eloyesp:
When instantiating a simple form for other ways that simple_form_for will
not set the builder correctly. This makes sure that the builder option is
always set.

I found this while testing a partial that receives a form builder instance, I've created that with `SimpleForm::FormBuilder.new`, but it behave a little bit different as the fields_for does not inherit the builder.

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -38,6 +38,7 @@ module SimpleForm
 
     def initialize(*) #:nodoc:
       super
+      options[:builder] ||= self.class
       @defaults = options[:defaults]
       @wrapper  = SimpleForm.wrapper(options[:wrapper] || SimpleForm.default_wrapper)
     end
--- a/test/action_view_extensions/builder_test.rb
+++ b/test/action_view_extensions/builder_test.rb
@@ -566,6 +566,13 @@ class BuilderTest < ActionView::TestCase
     end
   end
 
+  test 'fields for inherits the form builder class' do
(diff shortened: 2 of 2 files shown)


lucasmazza:
@eloyesp what kind of failure do you get without this patch? Would be nice to have a test case for it so we can track any regressions for this.

eloyesp:
@lucasmazza Added the test.

My issue was that when the form builder is instantiated without the simple_form_for method (using new) I've got an incomplete instance (without the builder option set), that behave different (as fields_for was not yielding a SimpleForm::FormBuilder instance in that case).

rafaelfranca:
One question to ask is: why it works if you use `simple_form_for`? Should we move something from the `simple_form_for` call to inside the initialize?

lucasmazza:
> One question to ask is: why it works if you use simple_form_for? Should we move something from the simple_form_for call to inside the initialize?

The view helpers ensure that `options[:builder]` is set, while instantiating a builder directly doesn't. This doesn't seem to be a common use case scenario but it's a patch we can live with.

eloyesp:
Any news about this PR?

feliperenan:
@rafaelfranca I think we can merge this PR. Even it isn't a common use case, the code is simpler and we can live with as Mazza said. Thoughts?

feliperenan:
To be honest, I don't know if we should merge it since it has been while and no one else asked for this patch. I'm gonna close this PR for now, but if someone else thinks that we should include this patch at Simple Form, please let us know.

Thanks

PR #1398, opened 2016-07-01
Use Rails 5 Attributes API for decimal_or_float?

olivierlacan:
The motivation for this commit originated in #1342.

@rafaelfranca mentioned that we should be using the new `type_for_attribute` interface
to determine the precise type of a column instead. At first I used the interface directly like
this:

```ruby
      def decimal_or_float?
        if object.respond_to?(:type_for_attribute)
          type = object.type_for_attribute(column.name.to_s).type
          type == :float || :decimal
        else
          column.number? && column.type != :integer
        end
      end
```

I then realized that we're already asking the column object directly for the type, so
why not simple check for the two types we care about and implement the method in a much
more straightforward *and* backward-compatible way? I ended up with this:

```ruby
      def decimal_or_float?
        column.type == :float || :decimal
      end
```

I could be missing something obvious but this is green in my Rails 5.0.0 app. The only
remaining issue I see is how the simple_form specs mock out the now removed number?
ActiveModel method: https://github.com/plataformatec/simple_form/blob/1c389f65594db90a1244d93a00a388b4aea4bcc6/test/support/models.rb#L3-L8

I'm gonna have to remove that most likely.

/cc @lucasmazza

### DIFF
--- a/lib/simple_form/inputs/base.rb
+++ b/lib/simple_form/inputs/base.rb
@@ -115,7 +115,7 @@ module SimpleForm
       end
 
       def decimal_or_float?
-        column.number? && column.type != :integer
+        column.type == :float || column.type == :decimal
       end
 
       def nested_boolean_style?
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ gemfile:
   - gemfiles/Gemfile.rails-4-0-stable
   - gemfiles/Gemfile.rails-4-1-stable
   - gemfiles/Gemfile.rails-4-2-stable
+  - gemfiles/Gemfile.rails-5-0-stable
   - Gemfile
 matrix:
   allow_failures:
--- a/gemfiles/Gemfile.rails-5-0-stable
+++ b/gemfiles/Gemfile.rails-5-0-stable
@@ -1,0 +1,15 @@
+source 'https://rubygems.org'
+
+gemspec :path => '..'
+
+platforms :rbx do
+  gem 'rubysl', '~> 2.0'
+  gem 'rubinius-developer_tools'
+end
+
+gem 'country_select', '~> 1.1.1'
+gem 'railties', github: 'rails/rails', branch: '5-0-stable'
+gem 'activemodel', github: 'rails/rails', branch: '5-0-stable'
+gem 'actionpack', github: 'rails/rails', branch: '5-0-stable'
+gem 'rake'
+gem 'tzinfo'
--- a/test/support/models.rb
+++ b/test/support/models.rb
@@ -1,10 +1,6 @@
 Association = Struct.new(:klass, :name, :macro, :scope, :options)
 
 Column = Struct.new(:name, :type, :limit) do
-  # Returns +true+ if the column is either of type integer, float or decimal.
-  def number?
-    type == :integer || type == :float || type == :decimal
-  end
 end
 
 Relation = Struct.new(:records) do


rafaelfranca:
My point is, we should not ask to the column anything anymore. The object that should be injected to the input should be an instance of `ActiveModel::Type::Value` in the versions that support it.

rafaelfranca:
This should be the method that should be changed https://github.com/plataformatec/simple_form/blob/81b9d7d3439225c7613e570afb44596acfa97a76/lib/simple_form/form_builder.rb#L551-L555

olivierlacan:
@rafaelfranca Makes sense. I'll update.

olivierlacan:
@rafaelfranca There's only one major issue though. My current PR is backward-compatible and easily releasable today.

What you're asking for will require some pretty major work to keep things backward-compatible, or will require a major version bump. No?

rafaelfranca:
Is the structure of the object that will be returned by `type_for_attribute` so different of what we have today? If not we can just call the methods conditionally. If it is we can create an adapter class to make the API the same.

lucasmazza:
> Is the structure of the object that will be returned by type_for_attribute so different of what we have today? If not we can just call the methods conditionally. If it is we can create an adapter class to make the API the same.

I think we only call `type` and `limit` on the column object, so using a type object instead should be OK - every implementation of `ActiveModel::Type::Value` has a `limit` attribute and a `type` method on it - but I guess we should upgrade our test suite to use ActiveModel types instead our column-like structs to confirm that.

lucasmazza:
Only API difference I noticed then playing on the console - `column_for_attribute` accepts Symbols, `type_for_attribute` doesn't.

```ruby
puts User.type_for_attribute('email').type #=> :string
puts User.column_for_attribute('email').type #=> :string

puts User.type_for_attribute('missing').type #=> nil
puts User.column_for_attribute('missing').type #=> nil

puts User.type_for_attribute(:email).type #=> nil
puts User.column_for_attribute(:email).type #=> :string
```

olivierlacan:
@lucasmazza Right so that felt a little overbearing for me to jump in and refactor a bunch of stuff like that. I can follow these instructions if you want though and we can review them together to see if they make sense. That said what about backward-compatibility? I think there's definitely more work necessary to create an adapter class as @rafaelfranca mentioned.

To me that sounds a lot like new minor release whereas with a somewhat lazy patch like the one I suggest in this PR's original commit, you have simple_form working with Rails 5 and prior versions today and we can take the time to make the gem compatible with old Rails versions and the new API while updating the tests to account for both.

> Only API difference I noticed then playing on the console - column_for_attribute accepts Symbols, type_for_attribute doesn't.

Yep, that originally confused me quite a bit.

sgrif:
Yeah, `type_for_attribute` intentionally lowers the API surface to only Strings. It's not meant to be called from app code directly, and we don't want HWIA all over our internals. As others have mentioned, Type objects are intentionally designed to have a somewhat similar API to what the columns used to have, but it may vary slightly.

lib/simple_form/inputs/base.rb
```diff
@@ -115,7 +115,7 @@ def decimal_limit
       end
 
       def decimal_or_float?
-        column.number? && column.type != :integer
+        column.type == (:float || :decimal)
```
BitOfUniverse: This is not the same as: `column.type == :float || column.type == :decimal`

olivierlacan: Good point, my bad.

PR #1402, opened 2016-07-12
only include ActionViewExtensions when ActionView is loaded

koenpunt:
this resolves a load order issue which I encountered when testing with rspec.

Also, the same problem occurred with https://github.com/bootstrap-ruby/rails-bootstrap-forms, which I solved in a similar manner in https://github.com/bootstrap-ruby/rails-bootstrap-forms/pull/275

### DIFF
--- a/lib/simple_form/action_view_extensions/builder.rb
+++ b/lib/simple_form/action_view_extensions/builder.rb
@@ -30,6 +30,8 @@ module SimpleForm
   end
 end
 
-ActiveSupport.on_load(:action_view) do
-  ActionView::Helpers::FormBuilder.send :include, SimpleForm::ActionViewExtensions::Builder
+module ActionView::Helpers
+  class FormBuilder
+    include SimpleForm::ActionViewExtensions::Builder
+  end
 end
--- a/lib/simple_form/action_view_extensions/builder.rb
+++ b/lib/simple_form/action_view_extensions/builder.rb
@@ -30,8 +30,6 @@ module SimpleForm
   end
 end
 
-module ActionView::Helpers
-  class FormBuilder
-    include SimpleForm::ActionViewExtensions::Builder
-  end
+ActiveSupport.on_load(:action_view) do
+  ActionView::Helpers::FormBuilder.send :include, SimpleForm::ActionViewExtensions::Builder
 end
--- a/lib/simple_form/action_view_extensions/form_helper.rb
+++ b/lib/simple_form/action_view_extensions/form_helper.rb
@@ -66,4 +66,6 @@ module SimpleForm
   end
 end
 
-ActionView::Base.send :include, SimpleForm::ActionViewExtensions::FormHelper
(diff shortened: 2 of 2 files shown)


koenpunt:
Turns out this is also part of an active discussion on rails-controller-testing: https://github.com/rails/rails-controller-testing/issues/24#issuecomment-231738446

lib/simple_form/action_view_extensions/builder.rb
```diff
@@ -30,8 +30,6 @@ def simple_fields_for(*args, &block)
   end
 end
 
-module ActionView::Helpers
-  class FormBuilder
-    include SimpleForm::ActionViewExtensions::Builder
-  end
+ActiveSupport.on_load(:action_view) do
```
rafaelfranca: This change seems not needed. On load hooks are only for constants that use it. The problem is to load action view base. Loading action view helpers and form builder should not cause any problem.

koenpunt: I've reverted that change, although the page here doesn't update..

koenpunt: You're right, removed it

koenpunt: I reverted that change, although github has some issues atm

rafaelfranca:
Awesome. Thank you for taking your time to fix a very spread bad practice that we have in the community. Even people that maintains Rails (like me) do this mistake.

koenpunt:
Not sure why Travis runs in to failures this second time though..  
  


On Jul 12 2016, at 6:13 pm, Rafael França &lt;notifications@github.com&gt;
wrote:  

> Awesome. Thank you for taking your time to fix a very spread bad practice
that we have in the community. Even people that maintains Rails (like me) do
this mistake.

>

> —  
> You are receiving this because you modified the open/close state.  
> Reply to this email directly, [view it on
GitHub](https://github.com/plataformatec/simple_form/pull/1402#issuecomment-232097524&r=cmVwbHkrMDAwNTViM2UxY2M4Nzk2NmMyNjg5YjcwZGJhNzQzY2Y1ZDUyZTM1YWI2ZTNmYzhkOTJjZjAwMDAwMDAxMTM5Y2Q5Mjc5MmExNjljZTA5ZDYyMjI5QHJlcGx5LmdpdGh1Yi5jb20=),
or [mute the
thread](https://github.com/notifications/unsubscribe/AAVbPrIubhRMeR3V44abhoG89vtrxqbjks5qU70ngaJpZM4JKLtr&r=cmVwbHkrMDAwNTViM2UxY2M4Nzk2NmMyNjg5YjcwZGJhNzQzY2Y1ZDUyZTM1YWI2ZTNmYzhkOTJjZjAwMDAwMDAxMTM5Y2Q5Mjc5MmExNjljZTA5ZDYyMjI5QHJlcGx5LmdpdGh1Yi5jb20=).

PR #1404, opened 2016-07-13
Ignore the `:label_text` when rendering `label` tags.

### LINKED ISSUE #1314: SimpleForm leaks the formatting function into HTML
When overriding the `label_text` per input like so:

```
form.label :name, label_text: proc { |label_text, required, _| "#{label_text} #{required}" }
```

The generated HTML has:

```
<label label_text="#<Proc:0x007fe268b21ce8@/Users/foo/boo/_new_form.html.slim:4>" class="string required" for="user_attributes_name">Name <abbr title="required">*</abbr></label>
```

As you can see it includes the full path to the custom proc function location which is a security issue.

This happens with versions `3.1.0` and `3.1.1.`

\cc @dnagir

lucasmazza:
Closes #1314.

### DIFF
--- a/test/form_builder/label_test.rb
+++ b/test/form_builder/label_test.rb
@@ -120,4 +120,11 @@ class LabelTest < ActionView::TestCase
       assert_select 'label[for=user_time_zone]', 'Time Zone:'
     end
   end
+
+  test 'builder allows label specific `label_text` option' do
+    with_label_for @user, :time_zone, label_text: lambda { |l, _, _| "#{l.titleize}:" }
+
+    assert_no_select 'label[label_text]'
+    assert_select 'label[for=user_time_zone]', 'Time Zone:'
+  end
 end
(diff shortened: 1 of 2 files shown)


lib/simple_form/components/labels.rb
```diff
@@ -50,7 +53,7 @@ def label_html_options
           label_options[:for] = options[:input_html][:id]
         end
 
-        label_options
+        label_options.except(*VIRTUAL_OPTIONS)
```
rafaelfranca: Did you see why it is here in the first place? `html_options_for(:label)`  should only return things that are in the `label_html` key.

rafaelfranca: Oh, we should be filtering it here https://github.com/plataformatec/simple_form/blob/81b9d7d3439225c7613e570afb44596acfa97a76/lib/simple_form/form_builder.rb#L302

lucasmazza: :+1:

test/form_builder/label_test.rb
```diff
@@ -120,4 +120,12 @@ def with_label_for(object, *args, &block)
       assert_select 'label[for=user_time_zone]', 'Time Zone:'
     end
   end
+
+
```
rafaelfranca: ✂️

rafaelfranca:
:shipit:

Issue #1380, opened 2016-05-06
Do not ignore implicitly passed required option


shyjt:
Simple form ignores implicitly passed required option when `browser_validations` is disabled. 
In simple form version 3 required option is always set to false for disabled `browser_validations` -  https://github.com/plataformatec/simple_form/blob/master/lib/simple_form/components/html5.rb
In version 2 it was able to pass required option even for disabled `browser_validations` - https://github.com/plataformatec/simple_form/blob/v2.2/lib/simple_form/components/html5.rb

I think explicitly passed required option should work for disabled `browser_validations`. What are your thoughts on this?

rafaelfranca:
Yeah, it makes sense. Could you send a PR  to fix it?

jdurand:
For reference, this worked as expected in v3.1.0.
This bug was introduced in v3.1.1.

rafaelfranca:
I think so.

lucasmazza:
Fixed by #1416

PR #1416, opened 2016-08-12
Do not ignore implicitly passed required option

petrgazarov:
Addresses https://github.com/plataformatec/simple_form/issues/1380

Confirmed that it was caused by https://github.com/plataformatec/simple_form/commit/a280ba770f2cbfa88b05e26cf83247d5b6900139. The breaking change was that it started to override the required option with `false` whenever `SimpleForm.browser_validations` was false.

Happy to add a spec if you agree with the approach 😄

### DIFF
--- a/lib/simple_form/components/html5.rb
+++ b/lib/simple_form/components/html5.rb
@@ -8,8 +8,8 @@ module SimpleForm
       def html5(wrapper_options = nil)
         @html5 = true
 
-        input_html_options[:required]        = has_required?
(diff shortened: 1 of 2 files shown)


lucasmazza:
> Happy to add a spec if you agree with the approach

:+1: Thanks for the PR, please go on with the tests 😄

petrgazarov:
Bump 😀

lucasmazza:
Thanks! :)

PR #1417, opened 2016-08-12
Support the ActiveModel::Type API when available

lucasmazza:
Another take to fix #1342, building on top of #1398.

We now favour the `ActiveModel::Type` API when available (Rails 5+), and still support the column objects from earlier versions. Our test models now implement the `type_for_attribute` API, but I discussed once with @rafaelfranca about adding an end to end test case based on ActiveRecord (+ sqlite) to provide better feedback on this integration.

### DIFF
--- a/test/form_builder/general_test.rb
+++ b/test/form_builder/general_test.rb
@@ -139,7 +139,11 @@ class FormBuilderTest < ActionView::TestCase
 
   test 'builder generates uuid fields for uuid columns' do
     with_form_for @user, :uuid
-    assert_select 'form input#user_uuid.string.uuid'
+    if defined? ActiveModel::Type
+      assert_select 'form input#user_uuid.string.string'
+    else
+      assert_select 'form input#user_uuid.string.uuid'
+    end
   end
 
   test 'builder generates password fields for columns that matches password' do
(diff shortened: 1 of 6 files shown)


rafaelfranca:
Amazing. :shipit:

test/form_builder/general_test.rb
```diff
@@ -139,7 +139,11 @@ def with_custom_form_for(object, *args, &block)
 
   test 'builder generates uuid fields for uuid columns' do
     with_form_for @user, :uuid
-    assert_select 'form input#user_uuid.string.uuid'
+    if defined? ActiveModel::Type
+      assert_select 'form input#user_uuid.string.string'
```
rafaelfranca: Why this is not uuid?

lucasmazza: In the stub `type_for_attribute` implementation we are returning the `:string` type for `:uuid` columns, so Simple Form will generate the class name as `string` - as the class name is derived from the `type.type` method - production apps probably won't be affected by this because the type object will likely be a [`ActiveRecord::ConnectionAdapters::PostgreSQL::OID::Uuid`](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/oid/uuid.rb) one.

rafaelfranca: Ok, so it is only because ActiveModel::Type doesn't have uuid registered?

lucasmazza: Exactly.

rafaelfranca: :+1:

Issue #1350, opened 2016-01-27
Proper id generation from name + value for radio buttons


jarosluv:
If I want to use non-latin radio buttons values, but helper generates not proper id by removing all non-latin chars. I think it must transliterate non-latin symbols for generating unique ids for proper using of labels.

This code

```
= simple_form_for User.new do |f|
  = f.input :name, as: :radio_buttons, collection: [ ["Технический писатель", "тексты"], ["Программист", "программы"] ]
```

will produce this html:

```
<form novalidate="novalidate" class="simple_form /users" accept-charset="UTF-8" method="post”>
  <div class="input radio_buttons required user_name”> 
    <label for="user_name_”>
      <input class="radio_buttons required" type="radio" value="тексты" name="user[name]" id="user_name_">Технический писатель
    </label>

    <label for="user_name_">
      <input class="radio_buttons required" type="radio" value="программы" name=“user[name]" id="user_name_">Программист
    </label>
  </div>
</form>
```

All labels have not proper references for according inputs.

feliperenan:
Duplicated https://github.com/plataformatec/simple_form/issues/1347

Issue #1347, opened 2016-01-25
using multibytes in collection breaks label's for and input's id attributes


github0013:
``` bash
$ uname -a
Darwin mbp 14.5.0 Darwin Kernel Version 14.5.0: Tue Sep  1 21:23:09 PDT 2015; root:xnu-2782.50.1~1/RELEASE_X86_64 x86_64

$ ruby -v
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin14]
```

created a repo to test.
https://github.com/github0013/simple_form_check_boxes

``` ruby
# app/views/users/create.html.erb

<%= simple_form_for User.new do |f| %>
  <%= f.input :tags, as: :check_boxes, collection: %w[tag1 tag2 tag3] %>
  <%= f.input :tags, as: :check_boxes, collection: %w[タグ１ タグ２ タグ３] %>
  <%= f.input :tags, as: :check_boxes, collection: Hash[*(%w[タグ１ タグ２ タグ３]).zip(1..3).flatten] %>
  <%= f.input :tags, as: :check_boxes, collection: Hash[*(%w[タグ１ タグ２ タグ３]).zip(%w[タグ１ タグ２ タグ３]).flatten] %>
<% end %>

```

While `collection: %w[tag1 tag2 tag3]` works as expected, `collection: %w[タグ１ タグ２ タグ３]` doesn't work properly when clicking on labels.

If you check the source, you will see the label `for` and input `id` are all same.  They are missing suffix part which should have been each of the collection text since the `%w[タグ１ タグ２ タグ３]` collection consists of **multibytes**.

Is there a way to fix this with some options?
**NOTE: tag texts itself should be sent to server, not some ids**

nashby:
this was fixed in Rails https://github.com/rails/rails/commit/d656097734928e1f6b3c3bd884e128c4e299f2ec

PR #1424, opened 2016-08-30
Add sanitized_options for collection check_boxes and radio_buttons

petrgazarov:
This PR aims to address a pain point for some users that try to use `check_boxes` or `radio_buttons` with `collection` array that contains non-latin characters. (https://github.com/plataformatec/simple_form/issues/1347 and  https://github.com/plataformatec/simple_form/issues/1350). The current code filters values through a latin regex (`/[^-\w]/`), and generates `id` and `for` attributes after that. This results in inaccurate values and broken functionality (cannot check checkbox by clicking on its label).

Here is a repo with working examples: https://github.com/petrgazarov/simple_form_check_boxes

Thanks for your work with simple_form btw, I'm a big fan and prefer it to vanilla forms every time!

### DIFF
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -1,6 +1,7 @@
 require 'action_view'
 require 'simple_form/action_view_extensions/form_helper'
 require 'simple_form/action_view_extensions/builder'
+require 'simple_form/action_view_extensions/tags'
 require 'active_support/core_ext/hash/slice'
 require 'active_support/core_ext/hash/except'
 require 'active_support/core_ext/hash/reverse_merge'
--- a/lib/simple_form/action_view_extensions/tags.rb
+++ b/lib/simple_form/action_view_extensions/tags.rb
@@ -1,0 +1,30 @@
+module ActionView
+  module Helpers
+    module Tags
+      class Base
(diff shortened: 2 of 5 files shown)


petrgazarov:
Any chance for a review here?

rafaelfranca:
Thank you for the pull request. While I agree this is a problem I don't think having a new option only for this is a good idea. Instead of `sanitized_options` I'd prefer that users provide the ids that they want to use.

Also, does this problem also exists in the Rails helpers? If it does, I think it is a better idea to fix this in Rails directly.

petrgazarov:
@rafaelfranca Yes, I had the same problem using rails forms. My thinking was that `simple_form` can provide this functionality as an extra above what rails form does. I do agree though, that it'd be even better if rails had that feature, which would naturally be included here.
Happy to explore the ids approach, otherwise feel free to close. Thanks!

ryanhageman:
Is this still an issue?

feliperenan:
I'm not sure with should go ahead with this PR since it hasn't recent activity. Can I close this one?

cc @rafaelfranca

feliperenan:
I'm close this PR since it hasn't recently activity.

mvz:
This is definitely still an issue.

PR #1435, opened 2016-10-12
Add bin/spring before rails to not hungs

Dorian:
Running `rails generate simple_form:install` hungs because of spring.


rafaelfranca:
Thank you for the pull request. If that is the case it is a problem in spring. Both things should work.

PR #1448, opened 2016-12-06
Make errors_on_attribute return [] when not present

### LINKED ISSUE #1447: undefined method `+' for nil:NilClass
My problem is that I have AR model, which has serialized attribute `data` and is serialized with a plain ruby object that includes `ActiveModel::Model` for validations.

what happens to me is that once object is submitted with validation error I am not able to load form again and I see:
```
undefined method `+' for nil:NilClass
```

which is pointing to particular line in code and I managed to get around it with small fix (will submit PR shortly)

the question here is if you can please help me find a solution here, or asses if what I did there is legit :)

Thank you :)

EDIT: 

forgot to mention that here is app to reproduce the problem: https://github.com/redrick/simple_form_error_demonstration

once you save user empty you will not be able to edit because of this mistake

redrick:
fixes #1447

### DIFF
--- a/lib/simple_form/components/errors.rb
+++ b/lib/simple_form/components/errors.rb
@@ -38,7 +38,7 @@ module SimpleForm
       end
 
       def errors_on_attribute
-        object.errors[attribute_name]
+        object.errors[attribute_name] || []
       end
 
       def full_errors_on_attribute


TiteiKo:
@rafaelfranca :up: this PR is simple and solves a bug occurring for a lot of folks, could it be included in the next release?

PR #1452, opened 2016-12-31
Use latest Ruby version in CI

takkanm:
I added test setting for Ruby2.4.
And, any Ruby version updated to latest version.


jcoyne:
It looks like Nokogiri 1.7.0 is not compatible with ruby < 2.2, so you'll have to pin nokogiri version in the Gemfile or drop those older rubies from the build matrix.

oliverklee:
Related: I've added PR #1455 to remove support for the unmaintained Ruby 1.9.x.

lucasmazza:
Closed by #1454

takkanm:
I see. Thank you !!

Issue #1320, opened 2015-11-11
unchecked: false does not work when `config.boolean_style = :inline`


ryanfelton:
Notice the in the code samples below how the hidden field is not generated when using the nested boolean style.

Code generated when using the default `config.boolean_style = :nested`

```
<div class="checkbox">
  <input value="false" type="hidden" name="meal[completed]">
  <label class="boolean optional" for="meal_completed">
  <input class="boolean optional" type="checkbox" value="true" name="meal[completed]" id="meal_completed">Completed</label>
</div>
```

Code generated when using the default `config.boolean_style = :inline`

```
<div class="checkbox">
  <input class="boolean optional" type="checkbox" value="true" checked="checked" name="meal[completed]" id="meal_completed">
  <label class="boolean optional" for="meal_completed">Completed</label>
</div>
```

Expected Code generated when using the default `config.boolean_style = :inline`

```
<div class="checkbox">
  <input value="false" type="hidden" name="meal[completed]">
  <input class="boolean optional" type="checkbox" value="true" checked="checked" name="meal[completed]" id="meal_completed">
  <label class="boolean optional" for="meal_completed">Completed</label>
</div>
```

ryanfelton:
Update: This bug only appears if you have custom value for `unchecked_value`. In this case `unchecked_value: false`

Example
`= f.input :completed, as: :boolean, checked_value: true, unchecked_value: false`

ryanfelton:
I've found a work around! 

Example
`= f.input :completed, as: :boolean, checked_value: true, unchecked_value: false`

Change the booleans from ruby `true` and `false` to strings `'true'` and `'false'`
`= f.input :completed, as: :boolean, checked_value: 'true', unchecked_value: 'false'`

dw2:
I'm experiencing the same issue without having custom checked/unchecked values. It's only when using `boolean_style: :inline` that it drops the unchecked value.

dw2:
My workaround is to put a hidden field before the checkbox to make sure the unchecked value is included:

```
= f.input :my_boolean_column, as: :hidden, input_html: { value: '0', id: '' }
= f.input :my_boolean_column
```

darrenterhune:
Wow banged my head for 2 hrs over this. Seems to be fixed in 2.1.3 the latest/last version for `rails 3` you can use. I was using `simple_form 2.1.1` where I was having this issue. Could probably close this @josevalim @rafaelfranca

fschwahn:
[...]

PR #1460, opened 2017-01-19
Exclude hidden field when unchecked_value: false

fschwahn:
Rails does not generate a hidden field for checkboxes if
`unchecked_value` is false. The nested boolean creates its own hidden
field and ignores this rule from rails. That means that the hidden
field is generated for a falsey value when using the nested style, but
not when using the inline style.

This commit fixes this problem by making the nested style behave like
the inline style, adhering to the standard set by rails.

Resolves: #1320

Note that this may cause problem for people relying in this faulty behavior. But I think it's a good idea to fix this inconsistency in the long run.

### DIFF
--- a/lib/simple_form/inputs/boolean_input.rb
+++ b/lib/simple_form/inputs/boolean_input.rb
@@ -60,7 +60,7 @@ module SimpleForm
       # we need the hidden field to be *outside* the label (otherwise it
       # generates invalid html - html5 only).
       def build_hidden_field_for_checkbox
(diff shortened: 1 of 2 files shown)


feliperenan:
Looks good to me. @rafaelfranca, any thoughts?

feliperenan:
Hi @fschwahn 

Could you rebase your branch with the master branch?

fschwahn:
@feliperenan thanks for moving forward with this, I just rebased.

Issue #1330, opened 2015-12-17
Using :full_error in wrapper causes empty divs for f.error and f.full_error


jeremysmithco:
I noticed that if I have a wrapper using :full_error instead of :error, and I then call f.error or f.full_error in my form, the form builder will add an empty div in the markup when first rendering the form.

If I switch back to using :error in my wrapper, calling f.error or f.full_error in the form does not add empty divs. (Empty divs wouldn't be a huge deal, except that I add a class to them with margins set, and that adds extra space when they are empty.)

Here's a how to reproduce it:

SimpleForm config:

```
SimpleForm.setup do |config|
  config.wrappers :default do |b|
    b.use :label_input
    b.use :full_error, wrap_with: { tag: :span }
  end

  config.default_wrapper = :default
end
```

A form which will render errors on base and the full_error on email:

```
<%= simple_form_for @user do |f| %>
  <%= f.error :base %>

  <%= f.input :email %>
  <%= f.full_error :email %>

  <%= f.button :submit %>
<% end %>
```

This will render the following markup:

```
<form accept-charset="UTF-8" action="/users" class="simple_form new_user" id="new_user" method="post">
  <div style="display:none">
    <input name="utf8" type="hidden" value="&#x2713;" /><input name="authenticity_token" type="hidden" value="fN+dzDMck0PT1MzL9DdVLVdjsVTrEXc7y2VvNq27hMo=" />
  </div>
  <div></div>

  <div class="email required user_email">
    <label class="email required" for="user_email"><abbr title="required">*</abbr> Email</label>
    <input class="string email required" id="user_email" name="user[email]" type="text" />
  </div>
  <div></div>

  <input class="button" name="commit" type="submit" value="Create User" />
</form>
```

If I switch the :default wrapper to `b.use :error`, the problem will go away. (And when errors are rendered, they will be within span tags, as specified in `:wrap_with`.)

feliperenan:
I don't know if it's an issue yet, so I'm closing it. Feel free to reopen it if necessary.

PR #1465, opened 2017-02-21
Add base error display alongside existing errors

bluefalcon26:
#1466

### DIFF
--- a/lib/generators/simple_form/templates/_form.html.erb
+++ b/lib/generators/simple_form/templates/_form.html.erb
@@ -1,5 +1,6 @@
 <%%= simple_form_for(@<%= singular_table_name %>) do |f| %>
-  <%%= f.error_notification %>
+	<%%= f.error_notification %>
+	<%%= f.error_notification message: f.object.errors[:base].to_sentence %>
 
   <div class="form-inputs">
   <%- attributes.each do |attribute| -%>
--- a/lib/generators/simple_form/templates/_form.html.erb
+++ b/lib/generators/simple_form/templates/_form.html.erb
@@ -1,6 +1,6 @@
 <%%= simple_form_for(@<%= singular_table_name %>) do |f| %>
-	<%%= f.error_notification %>
-	<%%= f.error_notification message: f.object.errors[:base].to_sentence %>
+  <%%= f.error_notification %>
+  <%%= f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present? %>
 
   <div class="form-inputs">
   <%- attributes.each do |attribute| -%>
--- a/lib/generators/simple_form/templates/_form.html.haml
+++ b/lib/generators/simple_form/templates/_form.html.haml
@@ -1,6 +1,6 @@
 = simple_form_for(@<%= singular_table_name %>) do |f|
   = f.error_notification
-  = f.error_notification message: f.object.errors[:base].to_sentence
+  = f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present?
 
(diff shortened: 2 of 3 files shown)


fabriziobertoglio1987:
Thanks a lot. Sorry, I am a beginner.. Is there somewhere an article on how to include this in my project master branch? 

Thanks a lot
Fabrizio

bluefalcon26:
@fabriziobertoglio1987 You should be able to do a git pull on 
https://github.com/bluefalcon26/simple_form.git:addBaseErrorDisplay

fabriziobertoglio1987:
Ok, thanks a lot. I will search for instruction on github to do a pull. I never have done one .. 

Thanks 
Fabrizio

lib/generators/simple_form/templates/_form.html.erb
```diff
@@ -1,5 +1,6 @@
 <%%= simple_form_for(@<%= singular_table_name %>) do |f| %>
   <%%= f.error_notification %>
+  <%%= f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present? %>
```
feliperenan: You can use `f.error :base` instead.

feliperenan:
@bluefalcon26 Thank you for the pull request 😄 . Could you rebase it with the master branch?

I don't see problems in adding it in the generator, @rafaelfranca any thoughts?

lib/generators/simple_form/templates/_form.html.erb
bluefalcon26: Does f.error :base automatically check whether an error is actually present?
Don't want to generate extra markup for no error message.

IE, Should the line say 
`<%%= f.error :base %>`
or
`<%%= f.error :base if ... %>`

See here for more info:
https://github.com/plataformatec/simple_form/issues/1330

feliperenan: > Does f.error :base automatically check whether an error is actually present?

Yes, it is.

> Don't want to generate extra markup for no error message.

According to my tests, it doesn't generate extra markup. The only difference that I saw was that with `f.error` uses `span` tag with class `error` instead `div` with class `error_notification`.

```ruby
f.error_notification
=> <div class="error_notification">Custom validation</div>

f.error
=> <span class="error">Custom validation</span>
```

collimarco:
Is this pull request the reason why there are two "duplicate" lines?

```
<%= f.error_notification %>
<%= f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present? %>
```

What is the reason for having this duplication?

bluefalcon26:
> Is this pull request the reason why there are two "duplicate" lines?
> 
> ```
> <%= f.error_notification %>
> <%= f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present? %>
> ```
> 
> What is the reason for having this duplication?

f.error_notification <> f.object.errors[:base]

The second line adds error details if they are present. In your case, are you getting error details in the first line, or are you getting two error lines without details?

collimarco:
@bluefalcon26 Thanks for the reply. Does `f.error_notification` alone display only a fixed message defined in YML (e.g. There are some errors) or it may include the specific runtime errors on some occasions? I'm tempted to remove the first `f.error_notification` since the errors are already displayed nicely below each field.

bluefalcon26:
Go ahead, it's possible things have changed since I was working with this
and it may be superceded. Only thing I would say is before you do, double
check the interaction on nested forms. It's been a long time but I seem to
recall that being part of the rationale for my original pull request.

Daniel

Issue #1475, opened 2017-03-28
Implement an :error_class attribute for inputs


CharlieHawker:
I've started to look at working on a Foundation 6 generator for SimpleForm, and one of the early stumbling blocks I've hit is the inability to set an error class on the actual input field; there is some chat (and workarounds) [over at StackOverflow](http://stackoverflow.com/questions/21559545/how-can-i-add-an-error-class-to-the-input-on-error-using-simple-form) but I don't see any open issues on the repository discussing this.

I believe this can be achieved fairly simply by adding the following code:

```
if wrapper_options.key?(:error_class) && has_errors?
  (wrapper_options[:class] ||= "").concat(" #{wrapper_options.delete(:error_class)}")
end
```

between [lines 192 and 193](https://github.com/plataformatec/simple_form/blob/master/lib/simple_form/inputs/base.rb#L192-L193) of `lib/simple_form/inputs/base.rb`.

The above seems to work for me and passes all tests, so I'll raise a PR.

remofritzsche:
I've tried your patch, and my inputs receive the :error_class even when there is no error (or no validation done at all). Debugging it I figured out that `has_errors?` in the context of `SimpleForm::Inputs::Base` appears to be `true` all the time in my case. Not sure if this is a general issue.

remofritzsche:
I've tried out your patch and just experienced something very strange: When having the server freshly started and I open the the form, it works as expected. When I submit the form, it correctly shows all the error classes. But when I reload the page, reopen the form and show it again, all of the inputs have the defined `:error_class` and `has_errors?` returns `true` - but the object `my_model.errors.present` is `false` (validation is not even called when just opening the form).

Looks like some very weird caching issue, probably in SimpleForm?

CharlieHawker:
Good spot, PR updated 👍

remofritzsche:
Great, thanks :) As I've just noticed, this solution works for `input`, but not for `input_field`. Do you have an idea how we could fix both?

CharlieHawker:
You're right, that's not ideal... I'll take a look and see if there's a nice solution.

remofritzsche:
Have you had a chance to look into this matter yet? That would be awesome.

CharlieHawker:
Hey @remofritzsche, will be taking a look either later today or tomorrow, sorry, I've been on other projects, and I made this modification with one in particular in mind 😄

remofritzsche:
[...]

PR #1476, opened 2017-03-28
Add ability to set an error_class directly on input

CharlieHawker:
Relates to issue #1475


remofritzsche:
Just wanted to create a PR this second for the exact same thing, but your solution is way more elegant. Would be wonderful if this gets merged and released :) Thank you for your work!

remofritzsche:
I've tried your patch and just found a bug: This code is adding a class to `wrapper_options[:class]`. This hash is shared between requests and statically cached, so changing this very instance creates adds the error class to all the inputs on second form reload and persists. 

A simple solution is to add `wrapper_options = wrapper_options.dup` right after the line `if wrapper_options`.

Now it works like a charm :)

CharlieHawker:
Good spot @remofritzsche, I'd been merrily using my branch without noticing that! 

I've implemented your suggestion, but with a slight amendment – I put it inside the second conditional block (the one that checks for the existence of the `:error_class` option) so as to not do the `wrapper_options.dup` unnecessarily.

airblade:
@CharlieHawker Doesn't this add the `error_class` to all inputs, even if only one has an error?

CharlieHawker:
@airblade it shouldn't do. @remofritzsche did raise an issue with the PR which I have since fixed (see earlier in the thread). Are you experiencing what he described?

remofritzsche:
I also couldn't notice this. I'm using the current patch for quite some time now and have never experienced any issues (though I'd love to see this for `input_field` as well).

anlek:
@CharlieHawker, Thanks for taking this on, seems like Bootstrap 4 beta could use this feature.

CharlieHawker:
@remofritzsche agreed, I need to sort that `input_field` bit out. Will try to do it this weekend.

Superpencil:
@CharlieHawker #pokethebear

m5o:
@CharlieHawker would you also add some tests, @rafaelfranca like them too 😬

yordis:
@rafaelfranca hey man, do you think we can have this merge as soon as you can?

printercu:
I've tried this patch, seems like it does work only in inline mode (`f.input :email, error_class: 'is-invalid'`) but does not work for wrappers defined in initializers with `b.use :input, error_class: 'is-invalid`. In this case it adds `error_class='is-invalid'` attr to valid fields, and when using with `class: 'form-control'` it expands this string with every new error (like this: `form-control is-invalid is-invalid is-invalid string email optional`), also making all other inputs `is-invalid`.

I've slightly modified it and implemented with patch like this:

```ruby
SimpleForm::Inputs::Base.prepend Module.new {
  def merge_wrapper_options(options, wrapper_options)
    if wrapper_options&.key?(:error_class)
      wrapper_options = wrapper_options.dup
      error_class = wrapper_options.delete(:error_class)
      wrapper_options[:class] = "#{wrapper_options[:class]} #{error_class}" if has_errors?
    end
    super(options, wrapper_options)
  end
}
```

feliperenan:
Hi @CharlieHawker 

Are you going to work on it yet?

m5o:
ℹ️ fyi, currently the `aria-required` attr. is added to an input with error.

https://github.com/plataformatec/simple_form/blob/7059550a5d841909f3fd3122f2b757ed2c9f1544/lib/simple_form/components/html5.rb#L9-L18

Here It would be easy to add additional class, don't know if this is a bad hack

```diff
input_html_options[:'aria-invalid']  = has_errors? || nil
+ input_html_options[:class] << 'is-invalid' if has_errors?
```

rafaelfranca:
That is definitely the wrong place 😄. I'm fine with this change with proper tests.

feliperenan:
I have continued this PR here #1552

feliperenan:
Closing in favor #1552

Issue #1462, opened 2017-02-07
How can simple_form be directed to produce :button instead of  :submit


jpwynn:
Is there a config setting that tells simple_form's generator to create forms with  

    = f.button :button   

instead of

    = f.button :submit    

In a Rails 5 app using simple_form for scaffolding, submit button are rendered like this:

    = f.button :submit 

which then gets rendered as  

    <input type="submit" ...

I ran into a problem when using a certain UI framework (materialize) the submit buttons only work if you click right on the text, but if you click off to the side the button css special effects activate but the button is not actually clicked.

rafaelfranca:
Thank you for the report.

See https://github.com/plataformatec/simple_form/pull/1482#issuecomment-293965259

PR #1482, opened 2017-04-13
Add ability to choose between f.button :submit and f.button :button when using scaffold generator

hubertjakubiak:
Resolves issue #1462


rafaelfranca:
Thank you for the pull request. The generators are only a convenience to quickly prototype. They would not work for every case and we don't plan to make them work for every case either. If users wants a button tag instead an input they could change the template in their applications.

PR #1483, opened 2017-04-13
Guess input type more carefully

sringling:
Match on word boundaries only.

The word "hourly" triggers a url type validation, that's not great.

### DIFF
--- a/test/form_builder/general_test.rb
+++ b/test/form_builder/general_test.rb
@@ -271,6 +271,11 @@ class FormBuilderTest < ActionView::TestCase
     assert_select 'form select#user_age.select'
   end
 
+  test 'builder does not generate url fields for columns that contain only the letters url' do
+    with_form_for @user, :hourly
+    assert_no_select 'form input#user_url.string.url'
+  end
+
   test 'builder allows overriding default input type for text' do
     with_form_for @user, :name, as: :text
     assert_no_select 'form input#user_name'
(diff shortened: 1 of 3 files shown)


test/form_builder/general_test.rb
```diff
@@ -217,6 +217,11 @@ def with_custom_form_for(object, *args, &block)
     assert_select 'form select#user_age.select'
   end
 
+  test 'builder does not generate url fields for columns that contain only the letters url' do
+    with_form_for @user, :hourly
+    assert_no_select 'form input#user_url.string.url'
```
feliperenan: Wdyt to add a assert to match the generated type?

sringling:
@feliperenan rebased!

test/form_builder/general_test.rb
rafaelfranca: Yeah, let's check the generated type.

feliperenan: @sringling Are you willing include this assert? after that we can merge it 😅

sringling: Not 100% sure I did what you meant, but I see the other tests tend to have these same 3 so I think I got it.

sringling:
@rafaelfranca updated for additional assert.

test/form_builder/general_test.rb
feliperenan: Now, it's ok for me 👍

feliperenan:
Hey @sringling, could you please rebase it again? 😅 after that we'll merge it.

sringling:
@feliperenan re-rebased!

PR #1484, opened 2017-04-24
Add citext input type

jazzque:
Just add support default pg datatype

### DIFF
--- a/lib/simple_form/inputs/citext_input.rb
+++ b/lib/simple_form/inputs/citext_input.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+module SimpleForm
+  module Inputs
+    class CitextInput < SimpleForm::Inputs::StringInput
+      def input(_wrapper_options = nil)
+        if email?
+          @builder.email_field(attribute_name, input_html_options)
+        else
+          @builder.text_field(attribute_name, input_html_options)
+        end
+      end
+
+      private
+
+      def email?
+        attribute_name == :email
+      end
+    end
+  end
+end
--- a/test/inputs/citext_input_test.rb
+++ b/test/inputs/citext_input_test.rb
@@ -1,0 +1,15 @@
+# encoding: UTF-8
+# frozen_string_literal: true
+require 'test_helper'
+
+class CitextInputTest < ActionView::TestCase
+  test 'input generates a email input type when email is citext' do
+    with_input_for @user, :email, :citext
+    assert_select "input.citext[type=email]"
+  end
+
+  test 'input generates a text area for citext' do
+    with_input_for @user, :name, :citext
+    assert_select "input.citext[type=text]"
+  end
+end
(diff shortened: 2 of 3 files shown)


rafaelfranca:
Thank you for the pull request but do we need an special class of input or just map citext datatype to the text input?

jazzque:
This is a default data type in pg. So except for adding this class to ur
project, I decided to make support this as default, like it was realized
'string with email = input type email'. And etc. Citext in the most of
cases using for email, less for another.
So, if u have this datatype in ur project, current realization generates an
invalid html with input type citext. So u have to add this class anyway. Or
use something like crutch in the config to map it.
Now this is flexible, easy to read, easy to support and supported by
default, so u don't need to wasting time for find-read-write ur own input
or using weird mapping in config. This just a simple extension.
This is just my opinion, thx for contact. Sorry for my English, I hope u
understand me. Have a good day!

On Apr 29, 2017 6:54 PM, "Rafael França" <notifications@github.com> wrote:

Thank you for the pull request but do we need an special class of input or
just map citext datatype to the text input?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/plataformatec/simple_form/pull/1484#issuecomment-298180539>,
or mute the thread

swrobel:
I also use citext for name fields so they're ordered case-insensitive, and also for unique constraints.

lib/simple_form/inputs/citext_input.rb
```diff
@@ -0,0 +1,19 @@
+module SimpleForm
```
feliperenan: Since we are using Ruby 2.4, could you add the special comment: `# frozen_string_literal: true`?

test/inputs/citext_input_test.rb
```diff
@@ -0,0 +1,14 @@
+# encoding: UTF-8
```
feliperenan: Since we are using Ruby 2.4, could you add the special comment: `#frozen_string_literal: true`?

rafaelfranca:
I still not sure if this is useful. This is just adding a class. Maybe we should be mapping the database type citext to the email input type?

swrobel:
@rafaelfranca [my PR for formtastic](https://github.com/justinfrench/formtastic/pull/1229) was considerably simpler but I believe the way simple_form is built requires these classes, unless you know of an easier way.

rafaelfranca:
We have similar code here https://github.com/plataformatec/simple_form/blob/master/lib/simple_form/form_builder.rb#L529

jazzque:
`I still not sure if this is useful. This is just adding a class. Maybe we should be mapping the database type citext to the email input type?`
follow by ur logic we dont even need any of persisted inputs in /lib/simple_form/inputs, coz we have a mapping which adds necessary input type

lib/simple_form/inputs/citext_input.rb
jazzque: @feliperenan fixd

test/inputs/citext_input_test.rb
jazzque: @feliperenan fixed

rafaelfranca:
You are not getting what I'm saying. We do need a mapping, we just don't need a new input for this. The email input can be used for the citext field.

PR #1486, opened 2017-04-28
Allow to use with Rails 5.1

### LINKED ISSUE #1485: Rails 5.1: Relax version spec in gemspec
Rails 5.1 is now out. SimpleForm requires Rails < 5.1.

```
simple_form (>= 3.4.0) was resolved to 3.4.0, which depends on
      actionpack (< 5.1, > 4)
```

Relax the gem version requirements?

doits:
fixes #1485

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ gemfile:
   - gemfiles/Gemfile.rails-4-1-stable
   - gemfiles/Gemfile.rails-4-2-stable
   - gemfiles/Gemfile.rails-5-0-stable
+  - gemfiles/Gemfile.rails-5-1-stable
   - Gemfile
 matrix:
   allow_failures:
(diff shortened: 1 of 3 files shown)


doits:
Used it successfully with rails 5.1.rc2 (simple form from rubygems).

Looks like the specs need some love now, though?

aried3r:
I think you need to adapt the Gemfile and regenerate the Gemfile.lock as well.

doits:
there are some other PR working on the test failures (#1487, #1488), so closing this one since I will not have time to investigate in near future

reza-naq:
@doits
>> ... since I will not have time to investigate in near future
Would it be right to say that 'simple_form' maintenance is not any more among the priorities for Plataformatec (now that the focus is more on Elixir)? If this is the case those in the community using ruby-based Plataformatec tools (Devise, Simple_forme) need to make a choice.

New version of rails is out and we can't upgrade to it because of the failures in the test suites of simple_forme. And there're many of us.

Here's a response I got from @dhh minutes ago:

> You hitting the dreaded upper-bound deadlock scenario. It's because simple_form has a pessimistic upper-bound of less-than Rails 5.1. So it won't resolve when you try to upgrade to Rails 5.1 because of that. You can try to upgrade simple_form and see if they have a version out that lifts the pessimistic restriction, but otherwise you'll have to fork the gem and remove that restriction. Yes, that's totally ridiculous. 
> We are currently working on a broad community response to avoid such needless deadlocking. Hope to have something ready soon. Sorry about this mess!

Thanks in advance

doits:
@reza-naq I think you got confused that I am an official Plataformatec member. There are other PRs officially working on Rails 5.1 support, so I Rails 5.1 support coming.

xanview:
Is there a branch or fork that works with 5.1? - I'm hitting this deadlock scenario :(

reza-naq:
The new release solves the problem. Thanks to the team.

PR #1487, opened 2017-04-28
[WIP] Support Rails 5.1

christiannelson:
Simple Form's gemspec prevents its use with Rails 5.1. Relaxing the version constraints leads to a great number of test failures. Luckily, the failures fall into one of three categories.

1. Missing polymorphic_mappings on MockController.
2. Type :text has been removed from ActiveModel (see https://github.com/rails/rails/commit/ef76f83f4cf0f27e84c0c5f4a3ff426d7ad84d9d)
3. TBD

Refers to https://github.com/plataformatec/simple_form/issues/1485.

Also looks like there's another PR for this (https://github.com/plataformatec/simple_form/pull/1486). Neither is complete though.


leifcr:
@christiannelson 

Please change the exclude part to these lines to the matrix in .travis.yml, as you only want tot test on ruby 2.4+ on rails 5.1

```yaml
matrix:
  allow_failures:
    - rvm: ruby-head
  exclude:
    - rvm: 2.0.0
      gemfile: Gemfile
    - rvm: 2.1.10
      gemfile: Gemfile
    - rvm: 2.4.0
      gemfile: gemfiles/Gemfile.rails-4-1-stable # activesupport 4.1 depends on json 1.8.3 which does not compile with Ruby 2.4.0
    - rvm: ruby-head
      gemfile: gemfiles/Gemfile.rails-4-1-stable # activesupport 4.1 depends on json 1.8.3 which does not compile with Ruby 2.4.0
    - rvm: jruby
      gemfile: Gemfile
    - rvm: 1.9.3
      gemfile: gemfiles/Gemfile.rails-5-0-stable
    - rvm: 2.0.0
      gemfile: gemfiles/Gemfile.rails-5-0-stable
    - rvm: 2.1.10
      gemfile: gemfiles/Gemfile.rails-5-0-stable
    - rvm: 2.2.7
      gemfile: gemfiles/Gemfile.rails-5-0-stable
    - rvm: jruby
      gemfile: gemfiles/Gemfile.rails-5-0-stable
    - rvm: 1.9.3
      gemfile: gemfiles/Gemfile.rails-5-1-stable
    - rvm: 2.0.0
      gemfile: gemfiles/Gemfile.rails-5-1-stable
    - rvm: 2.1.10
      gemfile: gemfiles/Gemfile.rails-5-1-stable
    - rvm: 2.2.7
      gemfile: gemfiles/Gemfile.rails-5-1-stable
    - rvm: 2.3.4
      gemfile: gemfiles/Gemfile.rails-5-1-stable
    - rvm: jruby
      gemfile: gemfiles/Gemfile.rails-5-1-stable
```

christiannelson:
@leifcr Thanks! I tweaked your suggestion a little, as Rails 5.0 and 5.1 should run on Ruby 2.2 and newer.

lukeasrodgers:
The fact that `:text` has been moved from AM to AR seems like it's going to make a fix a bit difficult. 

One approach would be to make active record a dependency, but then we'd still require a db connection in the simple_form tests in order for the call to `current_adapter_name` in AR `type.rb` to work. This seems too heavyweight. I don't think AR should be dependency of this gem and we also probably shouldn't require a db connection to run tests. 

This leads me to wonder if the right approach would be to delete the test that's failing and/or replace it with something higher level.

From #1417 it looks like the motivation for the currently-failing test was the introduction of this code  `if @object.respond_to?(:type_for_attribute) && @object.has_attribute?(attribute_name)` line in `form_builder.rb`. AFAICT this will continue to work just fine with rails 5.1.

lucasmazza:
Thanks for the PR! We went with #1488 instead and released Simple Form 3.5.0 😄

PR #1488, opened 2017-04-28
Adjust MockController & test support for Rails 5.1

rab:
* MockController#polymorphic_mappings
* Database-specific types (e.g., :text) moved from ActiveModel to
  ActiveRecord so the type_for_attribute helper needed adjustment.

Initially, this is just to allow testing with a client's system being updated to Rails 5.1.0


rab:
This might still have problems with ActiveRecord as it needs an adapter specified to understand the "meaning" of the `:text` type which was moved from ActiveModel to ActiveRecord.

test/support/models.rb
```diff
@@ -128,6 +128,7 @@ def column_for_attribute(attribute)
 
   begin
     require 'active_model/type'
+    require 'active_record/type'
```
matthewrudy: I think you might need to `require 'active_record/base'` to make this work

rab: Well, it worked without having to explicitly require it, but I'm removing the dependency on ActiveRecord anyway.

rab:
I've taken out the references to ActiveRecord that I'd added to take care of the `:text` type and split the test for the `description` attribute into an implicit version that expects to find an `input#user_description.string` and one with an explicit `as: :text` options expecting to find a `textarea#user_description.text`. I'm not sure that's really the best way, but if it is sufficient, there's probably also some minor documentation that needs to exist to explain why `as: :text` might be required.

rab:
Well, that's 💩 . Since older Rails will quite happily deal with `:text` in ActiveModel, I'll have to do something different for the tests.

rab:
I know this is ugly and will need to be squashed at the very least, but maintaining behavior across so many Rails versions is hard.

test/support/models.rb
```diff
@@ -128,29 +128,58 @@ def column_for_attribute(attribute)
...
-        when 'home_picture'  then :string
-        when 'amount'        then :integer
-        when 'attempts'      then :integer
-        when 'action'        then :string
-        when 'credit_card'   then :string
-        when 'uuid'          then :string
-      end
 
-      ActiveModel::Type.lookup(column_type, limit: limit)
+    begin
```
aCandidMind: Why not create a Hash first that maps all the types and then only overwrite that one specific entry?

rab:
That's a great idea. _(rebased the commits to spackle over some of the ugliness)_

test/support/models.rb
matthewrudy: I was looking at this run on travis
https://travis-ci.org/plataformatec/simple_form/jobs/226925127

that failed for that reason.

serggl:
Are we all good this this now? Excited to see this merged

test/form_builder/general_test.rb
```diff
@@ -117,8 +117,19 @@ def with_custom_form_for(object, *args, &block)
     assert_select 'form input#user_name.string'
   end
 
-  test 'builder generates text areas for text columns' do
+  test 'builder generates text field for text columns' do
     with_form_for @user, :description
+    if @user.respond_to?(:type_for_attribute) &&
+       @user.type_for_attribute('description').type == :string
+      assert_select 'form input#user_description.string'
```
rafaelfranca: Does this means that in Rails 5.1 a text column will return a input field instead of a text area? If so, this is a huge breaking change and we should handle it here and always return a text area for text column instead of changing the test.

lucasmazza: I think that only our test stubs are using `:string` instead of `:text` because `ActiveModel::Type::Text` no longer exists. We could backport it though:

```ruby
    begin
      ActiveModel::Type.lookup(:text)
    rescue ArgumentError
      class ::ActiveModel::Type::Text < ActiveModel::Type::String
        def type; :text; end
      end
      ActiveModel::Type.register(:text, ActiveModel::Type::Text)
    end
```

lucasmazza: @rab can you update the Pull Request and try this?

rab: @rafaelfranca I think that this means that database-specific types rely on the cooperation of the database adapter to distinguish between `:text` and `:string` as they are both represented in Ruby by `String`. As the testing environment does not include any database adapter, it must rely on the mocking of some of the type information where ActiveRecord might differ from ActiveModel.

Supporting the current range of Rails (actually, ActiveRecord and ActiveModel) versions means that things are a bit fiddly now with the `:text` type and there might need to be guidance on manually indicating inputs which should be rendered with a `textarea` in projects that can no longer automatically distinguish `:string` from `:text`

lucasmazza: With the changes from 5e6d306 we can revert the changes on this test case

.travis.yml
```diff
@@ -37,6 +38,14 @@ matrix:
       gemfile: gemfiles/Gemfile.rails-5-0-stable
     - rvm: jruby
       gemfile: gemfiles/Gemfile.rails-5-0-stable
+    - rvm: jruby
+      gemfile: gemfiles/Gemfile.rails-5-1-stable # rack 2.0.1 requires Ruby version >= 2.2.2; any sol'n for jruby?
```
rafaelfranca: I think we can use jruby-9.1.9.0 instead.

matthewrudy:
Nice!

PR #1337, opened 2015-12-27
Bootstrap4

tordans:
This is a WIP status but all I can do with my limited knowledge of both bootstrap4 and simple form.

If this looks like a good idea, please take over and finish it.

The PR has basically two parts:
- changing the form generator so it works for bootstrap 4 (details inside the extended commit messages)
- adding generator feature so both version 3 and 4 can be used. This part takes ideas from https://github.com/plataformatec/simple_form/issues/1336

PR #1492, opened 2017-05-20
Update Foundation initializer inline form to v6

Odaeus:
The `hidden-for-small-up` class no longer exists in Foundation 6. Replaced by the `hide` class. However, it seems more appropriate to use `show-for-sr` so that screen readers can read the field label.


Odaeus:
I realised after submitting that the README states explicitly support for Foundation 5. Should I duplicate the template to make one for version 6?

feliperenan:
Hi @Odaeus, thanks for the pull-request.

We won't support a specific CSS framework in Simple Form anymore. See https://github.com/plataformatec/simple_form/pull/1337#issuecomment-350889176

PR #1506, opened 2017-08-03
Fix For Invisible Checkboxes

tres:


rafaelfranca:
Thank you for the pull request but checkboxes inside labels are perfectly valid HTML. If they are not showing in a specific browser it is a browser bug.

PR #1510, opened 2017-08-17
fixed typos and indentations for _form generators

aoozdemir:
As titled.

### DIFF
--- a/lib/generators/simple_form/templates/_form.html.erb
+++ b/lib/generators/simple_form/templates/_form.html.erb
@@ -2,9 +2,9 @@
   <%= f.error_notification %>
 
   <div class="form-inputs">
-  <%- attributes.each do |attribute| -%>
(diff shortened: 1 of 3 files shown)


rafaelfranca:
Thank you for the pull request but those are not types. They need to be that way to generate valid code.

aoozdemir:
That's interesting. Can you explain a bit more, esp. double %?

rafaelfranca:
`<%%=` in the template will generate `<%=` in the view. `<%` is executed at template evaluation time so `<% attribute.each` we want to execute at template evaluation time and not in the generated code. Because of this it is not indented like the generated code.

aoozdemir:
Oh, I understand it now. `<%%=` is for escaping the interpolation. Didn't know that was a valid erb. But why is slim's indentation different than others?

rafaelfranca:
There is no reason really, the result is the same.

aoozdemir:
Alright! Thanks for the clarification. #TIL

PR #1520, opened 2017-10-04
add note about enumerize to readme

3wille:


rafaelfranca:
Thank you for the pull request but simple form has no context about enumerize so I don't think we should include that in the README.

PR #1530, opened 2017-12-08
Support for citext, hstore, json & jsonb columns

swrobel:
These are all commonly-used Postgres types. Fixes #1484 

@rafaelfranca I added tests but I'm concerned that they aren't actually accomplishing what they're supposed to. I tried using `uuid` as an example and deleted [the mapping](https://github.com/plataformatec/simple_form/blob/v3.5.0/lib/simple_form/form_builder.rb#L22) for that from `form_builder.rb` but the tests for it still passed. I believe that [test/support/models.rb](https://github.com/plataformatec/simple_form/blob/master/test/support/models.rb) might be responsible for the tests passing rather than the actual mapping from the gem. I'm a bit confused. However, this does work correctly in my application where I was using the previous workaround, so there's that.

### DIFF
--- a/test/form_builder/general_test.rb
+++ b/test/form_builder/general_test.rb
@@ -142,6 +142,24 @@ class FormBuilderTest < ActionView::TestCase
     assert_select 'form input#user_description.string'
   end
 
+  test 'builder generates text areas for hstore columns' do
+    with_form_for @user, :hstore
+    assert_no_select 'form input#user_hstore.string'
+    assert_select 'form textarea#user_hstore.text'
+  end
+
+  test 'builder generates text areas for json columns' do
+    with_form_for @user, :json
+    assert_no_select 'form input#user_json.string'
+    assert_select 'form textarea#user_json.text'
+  end
+
+  test 'builder generates text areas for jsonb columns' do
+    with_form_for @user, :jsonb
+    assert_no_select 'form input#user_jsonb.string'
+    assert_select 'form textarea#user_jsonb.text'
+  end
+
   test 'builder generates a checkbox for boolean columns' do
     with_form_for @user, :active
     assert_select 'form input[type=checkbox]#user_active.boolean'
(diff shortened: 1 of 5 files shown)


rafaelfranca:
I'll take a look in the tests. They looks weird to me too. Since we are here can we also add logic to change `citext` columns with name `email` to be of the `email` type. That will cover #1484.

tyrauber:
What sort of test coverage is required? It'd be awesome to get this merged. There are other gems that would benefit from simple_form supporting these column types.

test/form_builder/general_test.rb
```diff
@@ -136,6 +136,24 @@ def with_custom_form_for(object, *args, &block)
...
+  test 'builder generates text areas for hstore columns' do
+    with_form_for @user, :hstore
+    assert_no_select 'form input#user_hstore.string'
+    assert_select 'form textarea#user_hstore.text'
+  end
+
+  test 'builder generates text areas for json columns' do
+    with_form_for @user, :json
+    assert_no_select 'form input#user_json.string'
+    assert_select 'form textarea#user_json.text'
```
feliperenan: The reason that the tests are failing is because in other rails version (I used Rails 4.0 of example) the form is generating this path: `form textarea#user_hstore.json` - I don't no why, I didn't investigate it 🙈.

Maybe we can use just  `form textarea#user_hstore` for these tests? 🤔 Wdty @rafaelfranca ?

rafaelfranca: It is only 4.0? At this point I'm more inclined to stop supporting anything that is not Rails 5.0+

feliperenan: I used Rails 5.1 in my tests. I will check if it's ok in the 5.0 version. If so, can we remove the older versions?.

rafaelfranca: :+1:

feliperenan:
hi @swrobel  about the build, you can rebase your branch with the branch master, probably it will fix that.

swrobel:
@feliperenan done!

feliperenan:
@rafaelfranca Have you some consideration about this PR?

feliperenan:
Hey @swrobel, can you please rebase it again 😅 ?

swrobel:
@feliperenan done!

PR #1534, opened 2018-01-07
Code cleanup

Fornacula:
Noticed some old hash syntax (:key => value) in my project's simple_form initializer and ended up with rubocopping the whole gem :)


feliperenan:
Hi @Fornacula 

The specs are breaking in some builds. Can you please take a look?

Btw, thanks for the pull-request ❤️

Fornacula:
No prob! Should be all good now.

PR #1543, opened 2018-02-07
Add support for Rails 5.2.

gobijan:


simple_form.gemspec
```diff
@@ -1,25 +1,24 @@
-# -*- encoding: utf-8 -*-
-$:.push File.expand_path("../lib", __FILE__)
-require "simple_form/version"
+$LOAD_PATH.push File.expand_path('../lib', __FILE__)
+require 'simple_form/version'
 
 Gem::Specification.new do |s|
-  s.name        = "simple_form"
+  s.name        = 'simple_form'
```
rafaelfranca: Can you avoid all those changes in the gemspec that are not to add support to Rails 5.2?

gobijan: Sure. I just used Rubocops autoformat here. Will revert the changes.

gobijan: @rafaelfranca done.

gobijan:
I did the changes as requested by @rafaelfranca. Can you check?

Gemfile
```diff
@@ -7,6 +7,7 @@ platforms :rbx do
   gem 'rubinius-developer_tools'
 end
 
-gem 'country_select', '~> 2.5.2'
-gem 'railties'
+gem 'activemodel', '~> 5.2.0.rc1'
```
feliperenan: Why we need to have it in this Gemfile as well?

swistaczek: Hey @gobijan, I think that you should remove that line

gobijan: Had this one in because otherwise I couldn't bundle update. You have a better solution?

09:43 $ bundle update
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies....
Bundler could not find compatible versions for gem "actionpack":
  In Gemfile:
    railties (~> 5.2.0.rc1) was resolved to 5.2.0.rc1, which depends on
      actionpack (= 5.2.0.rc1)

    simple_form was resolved to 3.5.1, which depends on
      actionpack (< 6, > 4)

Bundler could not find compatible versions for gem "activemodel":
  In Gemfile:
    simple_form was resolved to 3.5.1, which depends on
      activemodel (< 6, > 4)

Could not find gem 'activemodel (< 6, > 4)', which is required by gem 'simple_form', in any of the sources.

Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    simple_form was resolved to 3.5.1, which depends on
      activemodel (< 6, > 4) was resolved to 4.0.1, which depends on
        activesupport (= 4.0.1)

    railties (~> 5.2.0.rc1) was resolved to 5.2.0.rc1, which depends on
      activesupport (= 5.2.0.rc1)

feliperenan: You can run the bundle just for the specific version, something like:

```
BUNDLE_GEMFILE=$PWD/gemfiles/Gemfile.rails-5-2-stable bundle install
```

feliperenan: I'm not sure if we have to set the version of railties in our main Gemfile, the bundle update is failing because that.

@tegon @rafaelfranca some suggestions?

mrinterweb:
This is blocking my rails 5.2 upgrade.

Gemfile
```diff
@@ -7,6 +7,7 @@ platforms :rbx do
   gem 'rubinius-developer_tools'
 end
 
-gem 'country_select', '~> 2.5.2'
-gem 'railties'
+gem 'country_select', '~> 3.1.1'
+gem 'activemodel', '~> 5.2.0.rc2'
+gem 'railties', '~> 5.2.0.rc2'
```
mrinterweb: Now that rails 5.2 has been released, you should update these.

gobijan: Done :)

mrinterweb:
Gemfile.lock needs to be updated to rails 5.2 as well.

m5o:
🛎  @feliperenan / @rafaelfranca 🤓

feliperenan:
@gobijan What do you think about squash your commits? all of them is related to the same thing - Add support for Rails 5.2.

Here a great blog post: https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/

m5o:
:octocat: https://blog.github.com/2016-04-01-squash-your-commits/

just good to know for maintainers, github repos settings have this "squash and merge" button. 🙄
reduces the ping pong…

Gemfile
feliperenan: @gobijan I think we don't need to have this change. If you remove the `railties` version and run bundle install it will download the last version available that from now it's 5.2.

We need just to keep the Gemfile.rails-5-2-stable .

Could you do this change? then we can merge it 😄

gobijan:
Usually I recommend githubs "squash and merge" when merging the pull request into master as @m5o already said. 
What is preventing from merging into master? Just let me know and I will adjust if needed.

simple_form.gemspec
```diff
@@ -20,6 +20,6 @@ Gem::Specification.new do |s|
 
   s.rubyforge_project = "simple_form"
 
-  s.add_dependency('activemodel', '> 4', '< 5.2')
-  s.add_dependency('actionpack', '> 4', '< 5.2')
+  s.add_dependency('activemodel', '> 4', '< 6')
```
rafaelfranca: Let's remove the upper bound. Let it break, with Rails 6.0 or maybe it will not as we can see it happen with Rails 5.2, but that will not require us to make a new release just to change the upper bound in case it works out of the box.

Gemfile
```diff
@@ -7,6 +7,7 @@ platforms :rbx do
   gem 'rubinius-developer_tools'
 end
 
-gem 'country_select', '~> 2.5.2'
-gem 'railties'
+gem 'country_select', '~> 3.1.1'
+gem 'activemodel', '~> 5.2.0'
+gem 'railties', '~> 5.2.0'
```
rafaelfranca: We don't want to specify the version here. We want to run the latest.

gobijan:
Makes sense @rafaelfranca. Included your proposed changes.

jesperronn:
💪 Thanks to every participator on this PR, and special thanks to @feliperenan for remembering to make the release afterwards! Thanks!

PR #1546, opened 2018-02-15
Remove support from Rails 4.0, 4.1 and 4.2

feliperenan:
We won't support version less than Rails 5.0 anymore in Simple Form.


rafaelfranca:
Doesn't it have some code to support those versions?

feliperenan:
@rafaelfranca I found these codes.

https://github.com/plataformatec/simple_form/blob/d1a383bd6f73ff912efd147eac4465570e1d4f71/lib/simple_form/components/minlength.rb#L33

https://github.com/plataformatec/simple_form/blob/d1a383bd6f73ff912efd147eac4465570e1d4f71/test/form_builder/wrapper_test.rb#L267

https://github.com/plataformatec/simple_form/blob/d1a383bd6f73ff912efd147eac4465570e1d4f71/lib/simple_form/components/maxlength.rb#L33

Do you think we can remove this support as well in this PR?

rafaelfranca:
I think it is better.

feliperenan:
@rafaelfranca done!

.travis.yml
```diff
@@ -25,9 +22,9 @@ matrix:
     - rvm: 2.1.10
       gemfile: Gemfile
     - rvm: 2.4.0
```
tegon: Since you removed these `Gemfile`s, I think you can remove this exclude clauses altogether
```diff
- - rvm: 2.4.0
-    gemfile: gemfiles/Gemfile.rails-4-1-stable # activesupport 4.1 depends on json 1.8.3 which does not compile with Ruby 2.4.0
- - rvm: ruby-head
-    gemfile: gemfiles/Gemfile.rails-4-1-stable # activesupport 4.1 depends on json 1.8.3 which does not compile with Ruby 2.4.0

File: CHANGELOG.md [hunk 2]
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Remove Remove support from Rails 4.0, 4.1 and 4.2 [@feliperenan](https://github.com/feliperenan)
```
amandasposito: ✂️  one `Remove`

feliperenan: 😱

PR #1551, opened 2018-02-27
Update README and CHANGELOG about updates 3.5.1

feliperenan:
These updates are about using Simple Form with non Active Record objects.

More information: https://github.com/plataformatec/simple_form/issues/1549

### DIFF
--- a/README.md
+++ b/README.md
@@ -1049,6 +1049,76 @@ by passing the html5 option:
 <%= f.input :expires_at, as: :date, html5: true %>
 ```
 
+### Using non Active Record objects
+
+There are few ways to build forms with objects that don't inherit from Active Record, as 
+follow:
+
+You can include the module `ActiveModel::Model`.
+
+```ruby
+class User
+  include ActiveModel::Model
+
+  attr_accessor :id, :name
+end
+```
+
+If you are using Presenters or Decorators that inherit from `SimpleDelegator` you can delegate 
+it to the model.
+
+```ruby
+class UserPresenter < SimpleDelegator
+  # Without that, Simple Form will consider the user model as the object.
+  def to_model 
+    self 
+  end
+end
+```
+
+You can define all methods required by the helpers.
+
+```ruby
+class User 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1049,6 +1049,66 @@ by passing the html5 option:
...
+end
+```
+
+You can define all methods required by the helpers.
+
+```ruby
+class User 
+  extend ActiveModel::Naming
+
+  def id; end;
```
davidbrusius: We don't need the semicolon after the `end` ✂️. There are other points we have to fix this, take a look at them too. 😉

README.md
```diff
@@ -1049,6 +1049,66 @@ by passing the html5 option:
...
+end
+```
+
+If you are using Presenters or Decorators that inherit from `SimpleDelegator` you can delegate 
+it to the model.
+
+```ruby
+class UserPresenter < SimpleDelegator
+  # Without that, Simple Form will consider the user model as the object.
+  def to_model; self; end
```
davidbrusius: Just a suggestion:

```ruby
def to_model
  self
end
```

Would it be more readable if we go like this? 

I tend to use single line definition when the method body is empty, but it's just my preference.

README.md
```diff
@@ -1049,6 +1049,76 @@ by passing the html5 option:
...
+    id 
+  end
+
+  def persisted? 
+    false 
+  end
+end
+```
+
+If your object don't implement those methods, you must explicit it when you are building
```
tegon: you must explicit it **tell it** when you are building the form

README.md
```diff
@@ -1049,6 +1049,76 @@ by passing the html5 option:
...
+```
+
+If your object don't implement those methods, you must explicit it when you are building 
+the form.
+
+```ruby
+class User 
+  attr_accessor :id, :name
+
+  # The only method required for use f.submit helper.
```
tegon: The only method required **to** use **the** f.submit helper.

PR #1552, opened 2018-03-03
Allow custom errors classes to inputs

feliperenan:
Allow custom errors classes to input in our wrappers. With it, when a validation happens, the input will be rendered with the error class defined in the wrapper.

### Examples

We can configure an wrapper:
```ruby
  config.wrappers :form, tag: 'div', class: 'field', error_class: 'invalid' do |b|
    b.use :label

    b.use :input, error_class: 'is-invalid'
  end
```

And when a form with errors be rendered, will generated the follow markup:

```HTML
<div class="field invalid">
  <input class="string required is-invalid" type="text" value="ok" name="user[name]" id="user_name">
</div>
```

### Example working on Bootstrap 4

Thank you @CharlieHawker by #1476 and #1475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
-* Remove support from Rails 4.0, 4.1 and 4.2 [@feliperenan](https://github.com/feliperenan)
+* Allow custom errors classes to inputs . [@feliperenan](https://github.com/feliperenan)
+* Remove support from Rails 4.0, 4.1 and 4.2. [@feliperenan](https://github.com/feliperenan)
 
 ### Bug fix
 * Fix horizontal form label position, from right to text-right. [@cavpollo](https://github.com/cavpollo)
--- a/README.md
+++ b/README.md
@@ -901,11 +901,12 @@ You can customize _Form components_ passing options to them:
 
 ```ruby
 config.wrappers do |b|
-  b.use :label_input, class: 'label-input-class'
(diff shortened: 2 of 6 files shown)


m5o:
Awesome @feliperenan 
Do you think the positive case https://github.com/plataformatec/simple_form/pull/1333 can also land in the next release? Would be a perfect pair.

feliperenan:
Thanks @m5o ❤️ 

I think so. 

But #1333 adds the valid class just to the wrapper not to the input. After merging #1333 we'll need to have a code similar to the introduced on this PR to have the valid class in the input as well.

m5o:
> But #1333 adds the valid class just to the wrapper not to the input. After merging #1333 we'll need to have a code similar to the introduced on this PR to have the valid class in the input as well.

This would be awesome, looking forward for all these changes. Would like to include the new features to the bootstrap v4 demo https://github.com/rafaelfranca/simple_form-bootstrap/pull/56

remofritzsche:
Does this solution allow to set the error class on fields created with `input_field` as well? This would be very important for us. Thanks for your work 👍

feliperenan:
@remofritzsche I think it doesn't work with `input_field` yet, I will take a look on it.

remofritzsche:
I'd also welcome a release very much. It would be awesome though if support for custom error classes on fields created with `input_field` could also be included - we're depending heavily on that. Thanks a bunch for considering this 👍

PR #1553, opened 2018-03-07
Add :valid_class on input wrapper when value is present and valid v2

m5o:
This just continues the great work started by @aeberlin with #1333 

---

##### Purpose

The purpose of this PR is to improve style support for inputs with validations. Simply put, the lack of an applied `:error_class` is not indicative of the attribute being valid. While this isn't a guaranteed method of determining if it is valid, it is as close as programmatically possible (AFAIK). See caveats below.

##### Logic

When a `:valid_class` is specified, it will be applied to an input wrapper given the following conditions:

1. Object exists.
1. Object responds to `:errors`.
1. There are no validation errors (e.g. validations passed or were not specified).
1. Object responds to the attribute method.
1. Return value of attribute method is present.

##### Changes

1. Added `:valid_class` declarations (similar to those for `:error_class`) on default wrappers.
1. Added `:has_value?` and `:valid?` to `SimpleForm::Components::Errors`.
1. Adjusted `SimpleForm::Wrappers::Root#html_classes` to apply `:valid_class` when applicable.

##### Known caveats

If an attribute is valid when `:blank?`, the `:valid_class` will not be applied. I am open to suggestions here, as this is the only hole in my logic, albeit an acceptable and predictable one (IMO).

##### Further notes

I adjusted two failing tests that were missing stubs, and added another test for this feature.

Please let me know if there's something else I should include. Thanks, cheers!

### DIFF
--- a/lib/simple_form/wrappers/root.rb
+++ b/lib/simple_form/wrappers/root.rb
@@ -30,6 +30,7 @@ module SimpleForm
         end
         css << (options[:wrapper_error_class] || @defaults[:error_class]) if input.has_errors?
         css << (options[:wrapper_hint_class] || @defaults[:hint_class]) if input.has_hint?
+        css << (options[:wrapper_valid_class] || @defaults[:valid_class]) if input.valid?
         css.compact
       end
     end
--- a/test/form_builder/error_test.rb
+++ b/test/form_builder/error_test.rb
@@ -272,14 +272,4 @@ class ErrorTest < ActionView::TestCase
       assert_no_select 'span.error'
     end
   end
-
-  test 'input wrapper has valid class for present attribute without errors' do
-    @user.instance_eval do
-      undef errors
-      name = 'foo'
-    end
-
-    with_form_for @user, :name
-    assert_select '.input.field_without_errors'
-  end
 end
--- a/test/form_builder/error_test.rb
+++ b/test/form_builder/error_test.rb
@@ -273,8 +273,6 @@ class ErrorTest < ActionView::TestCase
     end
   end
 
-  # NO ERRORS
-
   test 'input wrapper has valid class for present attribute without errors' do
     @user.instance_eval do
       undef errors
--- a/test/form_builder/error_test.rb
+++ b/test/form_builder/error_test.rb
@@ -272,4 +272,16 @@ class ErrorTest < ActionView::TestCase
       assert_no_select 'span.error'
     end
   end
+
+  # NO ERRORS
+
+  test 'input wrapper has valid class for present attribute without errors' do
+    @user.instance_eval do
+      undef errors
+      name = 'foo'
+    end
+
+    with_form_for @user, :name
+    assert_select '.input.field_without_errors'
+  end
 end
--- a/test/form_builder/general_test.rb
+++ b/test/form_builder/general_test.rb
@@ -242,6 +242,7 @@ class FormBuilderTest < ActionView::TestCase
   test 'builder generates file for file columns' do
     @user.avatar = MiniTest::Mock.new
     @user.avatar.expect(:public_filename, true)
+    @user.avatar.expect(:!, false)
 
     with_form_for @user, :avatar
     assert_select 'form input#user_avatar.file'
@@ -250,6 +251,7 @@ class FormBuilderTest < ActionView::TestCase
   test 'builder generates file for attributes that are real db columns but have file methods' do
     @user.home_picture = MiniTest::Mock.new
     @user.home_picture.expect(:mounted_as, true)
+    @user.home_picture.expect(:!, false)
 
     with_form_for @user, :home_picture
     assert_select 'form input#user_home_picture.file'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Allow custom errors classes to inputs.[@feliperenan](https://github.com/feliperenan)
 * Remove support from Rails 4.0, 4.1 and 4.2. [@feliperenan](https://github.com/feliperenan)
 * Add support for citext, hstore, json & jsonb column types. [@swrobel](https://github.com/swrobel)
+* Add :valid_class on input wrapper when value is present and valid [@aeberlin](https://github.com/aeberlin), [@m5o](https://github.com/m5o)
 
 ### Bug fix
 * Fix horizontal form label position, from right to text-right. [@cavpollo](https://github.com/cavpollo)
--- a/README.md
+++ b/README.md
@@ -856,7 +856,8 @@ The syntax looks like this:
 
 ```ruby
 config.wrappers tag: :div, class: :input,
-                error_class: :field_with_errors do |b|
+                error_class: :field_with_errors,
+                valid_class: :field_without_errors do |b|
 
   # Form extensions
   b.use :html5
--- a/lib/generators/simple_form/templates/config/initializers/simple_form.rb
+++ b/lib/generators/simple_form/templates/config/initializers/simple_form.rb
@@ -14,7 +14,7 @@ SimpleForm.setup do |config|
   # stack. The options given below are used to wrap the
   # whole input.
   config.wrappers :default, class: :input,
-    hint_class: :field_with_hint, error_class: :field_with_errors do |b|
+    hint_class: :field_with_hint, error_class: :field_with_errors, valid_class: :field_without_errors do |b|
     ## Extensions enabled by default
     # Any of these extensions can be disabled for a
     # given input by passing: `f.input EXTENSION_NAME => false`.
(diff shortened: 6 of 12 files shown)


feliperenan:
> If an attribute is valid when :blank?, the :valid_class will not be applied. I am open to suggestions here, as this is the only hole in my logic, albeit an acceptable and predictable one (IMO).

To deal with it, maybe we can use [ActiveModel::Dirty](http://api.rubyonrails.org/classes/ActiveModel/Dirty.html). Something like:

```ruby
def changed?
  changed_method = "#{attribute_name}_changed?"
  return false unless object && object.respond_to?(changed_method)

  object.public_send(changed_method)
end
```

But it will work just for `ActiveRecord` models or objects that include the `ActiveModel::Dirty` module.

test/form_builder/error_test.rb
```diff
@@ -272,4 +272,14 @@ def with_full_error_for(object, *args)
       assert_no_select 'span.error'
     end
   end
+
+  test 'input wrapper has valid class for present attribute without errors' do
```
feliperenan: We could add wrapper test as well, like this one:

https://github.com/plataformatec/simple_form/blob/31720f00f9ce482e8c895fba6707cbdea08710cf/test/form_builder/wrapper_test.rb#L36

Wdty?

m5o: This test is currently only testing the wrapper div. I'll move it to the `wrapper_test.rb`

```html
<form class="simple_form edit_user" id="edit_user_1" action="/users" accept-charset="UTF-8" method="post"><input name="utf8" type="hidden" value="&#x2713;" /><input type="hidden" name="_method" value="patch" /><div class="input string required user_name field_without_errors"><label class="string required" for="user_name"><abbr title="required">*</abbr> Super User Name!</label><input class="string required" required="required" aria-required="true" maxlength="100" size="100" type="text" value="New in SimpleForm!" name="user[name]" id="user_name" /></div></form>
```

m5o:
Is there something more to add? Don't know how to act about the AM::Dirty suggestion.

rafaelfranca:
It is better to not add another dependency to use simple form. I'm fine with that known caveats as it is right now

lib/simple_form/wrappers/root.rb
```diff
@@ -30,6 +30,7 @@ def html_classes(input, options)
         end
         css << (options[:wrapper_error_class] || @defaults[:error_class]) if input.has_errors?
         css << (options[:wrapper_hint_class] || @defaults[:hint_class]) if input.has_hint?
+        css << (options[:wrapper_valid_class] || @defaults[:valid_class]) if input.valid?
```
rafaelfranca: Why we are using a new `valid?` method and not the `!has_errors`?

test/form_builder/general_test.rb
```diff
@@ -250,6 +251,7 @@ def with_custom_form_for(object, *args, &block)
   test 'builder generates file for attributes that are real db columns but have file methods' do
     @user.home_picture = MiniTest::Mock.new
     @user.home_picture.expect(:mounted_as, true)
+    @user.home_picture.expect(:!, false)
```
rafaelfranca: I'm confused with this. What is this expecting?

m5o: Without this the tests are failing:
<details>
<summary>test_builder_generates_file_for_attributes_that_are_real_db_columns_but_have_file_methods</summary>

```sh
FormBuilderTest#test_builder_generates_file_for_attributes_that_are_real_db_columns_but_have_file_methods:
NoMethodError: unmocked method :!, expected one of [:mounted_as]
    /mac/.rvm/gems/ruby-2.4.3@simple_form/gems/activesupport-5.1.0/lib/active_support/core_ext/object/blank.rb:17:in `blank?'
    /mac/.rvm/gems/ruby-2.4.3@simple_form/gems/activesupport-5.1.0/lib/active_support/core_ext/object/blank.rb:24:in `present?'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/components/errors.rb:18:in `has_value?'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/components/errors.rb:22:in `valid?'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/wrappers/root.rb:33:in `html_classes'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/wrappers/many.rb:59:in `wrap'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/wrappers/many.rb:33:in `render'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/wrappers/root.rb:16:in `render'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/form_builder.rb:123:in `input'
    /mac/code/private/github/m5o__simple_form/test/support/misc_helpers.rb:249:in `block in with_form_for'
    /mac/.rvm/gems/ruby-2.4.3@simple_form/gems/actionview-5.1.0/lib/action_view/helpers/capture_helper.rb:39:in `block in capture'
    /mac/.rvm/gems/ruby-2.4.3@simple_form/gems/actionview-5.1.0/lib/action_view/helpers/capture_helper.rb:203:in `with_output_buffer'
    /mac/.rvm/gems/ruby-2.4.3@simple_form/gems/actionview-5.1.0/lib/action_view/helpers/capture_helper.rb:39:in `capture'
    /mac/.rvm/gems/ruby-2.4.3@simple_form/gems/actionview-5.1.0/lib/action_view/helpers/form_helper.rb:450:in `form_for'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/action_view_extensions/form_helper.rb:27:in `block in simple_form_for'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/action_view_extensions/form_helper.rb:46:in `with_simple_form_field_error_proc'
    /mac/code/private/github/m5o__simple_form/lib/simple_form/action_view_extensions/form_helper.rb:26:in `simple_form_for'
    /mac/code/private/github/m5o__simple_form/test/support/misc_helpers.rb:232:in `with_concat_form_for'
    /mac/code/private/github/m5o__simple_form/test/support/misc_helpers.rb:248:in `with_form_for'
    /mac/code/private/github/m5o__simple_form/test/form_builder/general_test.rb:256:in `block in <class:FormBuilderTest>'
```
</details>

lib/simple_form/wrappers/root.rb
m5o: Sorry no idea, this code is untouched and original by @aeberlin

m5o: Only `!has_errors?` would add the `valid_class` to a new rendered form by default. Combine the two `!has_errors? && has_value?` adds the class only on form submit and if the value is present and valid.

m5o:
@rafaelfranca @feliperenan Do you think we are ready to go ahead to review?

feliperenan:
Is possible add this feature using custom components that are being introduced in #1556 ? :thinking:

I wonder how much this feature will be useful since this is a backend validation. For me, it makes a lot of sense when we are talking about adding feedbacks in client side to valid and invalid inputs since the user can see the feedbacks in real-time.

I think that few users will enjoy this feature, so I'd prefer to leave to themselves implement it case it is possible. Case not, we can go ahead and merge it.

What are your thoughts? @rafaelfranca @m5o

m5o:
I think this is a nice feature with adds value to `simple_form` out-of-the-box. Of curse, live validation is a different topic. But just think of long forms, or forms with inputs (like file upload), that required but often forgotten to fill. The visual feedback (valid or invalid) give the end-user guidance about what to change or skip (User don't read every content, they visually scan pages). I would like to [demo](https://simple-form-bootstrap4.herokuapp.com/) this, as soon as the custom inputs is merged to master, then i'll rebase and change the branch this.

If a user didn't want/like this feature, they can just reset the `:valid_class` in their initializer. The adoption is wide if we pick up nice defaults. I think actually only a few can implement this by themselves, thats why they pick this gem 😉 

I would volunteer to add this feature also to inputs, [we discussed days ago](https://github.com/plataformatec/simple_form/pull/1552#issuecomment-370153289) to prepare support for all kinds of css frameworks or user preferences. 💪

m5o:
@rafaelfranca Still good?
> [👍 I'm good with the idea](https://github.com/plataformatec/simple_form/pull/1333#issuecomment-370912095)

m5o:
https://github.com/plataformatec/simple_form/pull/1553/commits/2f71508d6bda2e4bc9ec24b70724989aa044f3f6 adds the possibility to add the `valid_class` to inputs. Technically it depends from the same origin checks. Hope it's ok for everyone not separate this into another pr ✌️

feliperenan:
I see your points about have this feature in Simple Form. I still think that few users will enjoy that, though I think we can go ahead and merge it since I don't know another way to do that easily.

@m5o I think is better separate this commit on another PR to explain the motivations and facilitate the code review. After you do that, you can merge this one 😄.

PR #1554, opened 2018-03-07
add InputGroup component

m5o:
This is a spin off and draft after @rafaelfranca's [comment](https://github.com/rafaelfranca/simple_form-bootstrap/pull/56#discussion_r172645536) on https://github.com/rafaelfranca/simple_form-bootstrap/pull/56

* add InputGroup component

Currently the [_form.html.erb](https://github.com/rafaelfranca/simple_form-bootstrap/pull/56/files#diff-f7c4112a5b12258f4d3de675692a6c2b) requires an custom `:input_group` wrapper

```rb
SimpleForm.setup do |config|
  # Input Group
  config.wrappers :input_group, tag: 'div', class: 'form-group', error_class: 'form-group-invalid' do |b|
    b.use :html5
    b.use :placeholder
    b.optional :maxlength
    b.optional :minlength
    b.optional :pattern
    b.optional :min_max
    b.optional :readonly
    b.use :label, class: 'form-control-label'
    b.wrapper :input_group_tag, tag: 'div', class: 'input-group' do |ba|
      ba.optional :prepend
      ba.use :input, class: 'form-control', error_class: 'is-invalid'
      ba.optional :append
    end
    b.use :full_error, wrap_with: { tag: 'div', class: 'invalid-feedback d-block' }
    b.use :hint, wrap_with: { tag: 'small', class: 'form-text text-muted' }
  end
end
```

I've initially thought because [Bootstrap Input Groups](https://getbootstrap.com/docs/4.0/components/input-group/) are a feature explicitly used by bootstrap a global simple_form component doesn't make sense. Let's discuss.

### DIFF
--- a/lib/simple_form/components/input_groups.rb
+++ b/lib/simple_form/components/input_groups.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+module SimpleForm
+  module Components
+    module InputGroups
+      module Prepends
+        def prepend(wrapper_options = nil)
+          template.content_tag(:div, content_tag(:span, options[:prepend], class: "input-group-text"), class: "input-group-prepend")
+        end
+      end
+
+      module Appends
+        def append(wrapper_options = nil)
+          template.content_tag(:div, content_tag(:span, options[:append], class: "input-group-text"), class: "input-group-append")
+        end
+      end
+    end
+  end
+end
--- a/lib/simple_form/components.rb
+++ b/lib/simple_form/components.rb
@@ -12,6 +12,7 @@ module SimpleForm
     autoload :Errors
     autoload :Hints
     autoload :HTML5
+    autoload :InputGroups
     autoload :LabelInput
     autoload :Labels
     autoload :MinMax


feliperenan:
What do you think for add documentation about how use it?

lib/simple_form/components/input_groups.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+module SimpleForm
+  module Components
+    module InputGroups
+      module Prepends
+        def prepend(wrapper_options = nil)
```
feliperenan: Could we add tests to it?

lib/simple_form/components/input_groups.rb
```diff
@@ -0,0 +1,18 @@
...
+          template.content_tag(:div, content_tag(:span, options[:prepend], class: "input-group-text"), class: "input-group-prepend")
+        end
+      end
+
+      module Appends
+        def append(wrapper_options = nil)
+          template.content_tag(:div, content_tag(:span, options[:append], class: "input-group-text"), class: "input-group-append")
```
feliperenan: To improve the readability, we can extract the span tag, like this:

```ruby
span_tag = content_tag(:span, options[:append], class: "input-group-text")
template.content_tag(:div, span_tag, class: "input-group-append")
```

Wdyt?

lib/simple_form/components/input_groups.rb
m5o: Would you help me with the test setup? There is no components test I could cheat on.

m5o:
> What do you think for add documentation about how use it?

ok, but it will be depend how and when the initializer with bootstrap v4 support will land. As I said, Input groups are a bootstrap feature and it require to load initializer and style config.

lib/simple_form/components/input_groups.rb
feliperenan: To use this feature we need to have a wrapper? How I build a input using it?

lib/simple_form/components/input_groups.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+module SimpleForm
+  module Components
+    module InputGroups
+      module Prepends
+        def prepend(wrapper_options = nil)
+          span_tag = content_tag(:span, options[:prepend], class: "input-group-text")
```
feliperenan: If the tags and the html classes are customizable it could work with another CSS frameworks as well. Makes sense? 🤔

m5o: Yes, totally agree! Do you have a suggestion how to prepare these css classes for customization?

rafaelfranca:
Do we need a wrapper for this. Can't we implement the input group using HTML?

Something like this:

```erb
<div class="input-group mb-3">
  <div class="input-group-prepend">
    <span class="input-group-text" id="basic-addon1">@</span>
  </div>
  <%= f.input username %>
</div>
```

rafaelfranca:
Just saw this is exactly how in implemented in Bootstrap 3. I don't think we need to change simple form to support this. It is more a visual organization of inputs, for that we should be using HTML, we don't need to generate all the HTML using simple  form.

m5o:
> Just saw this is exactly how in implemented in Bootstrap 3.

I'm confused. Are we target [bootstrap v3](https://getbootstrap.com/docs/3.3/css/#forms) or [boostrap v4](https://getbootstrap.com/docs/4.0/components/input-group/)?

feliperenan:
Thinking better, even if we let tags and classes customizable the markup still be specific to Bootstrap.

We can include this component as a tutorial in a Wiki Page with a link to [simple_form-boostrap](https://github.com/rafaelfranca/simple_form-bootstrap). What do you think @m5o and @rafaelfranca ?

m5o:
I've prepared the [InputGroups definition](https://github.com/m5o/simple_form-bootstrap/blob/4008b0f08aabd000e0ddff537962dc225c87744a/config/initializers/simple_form_bootstrap.rb#L384-L404) and the [`:input_group` wrapper](https://github.com/m5o/simple_form-bootstrap/blob/4008b0f08aabd000e0ddff537962dc225c87744a/config/initializers/simple_form_bootstrap.rb#L308-L325) inside a initializer. 
Is this a no-go or are you guys ok with this solution? 

I prefer to ship solid default, maybe disabled/[commented] but working instead of forgotten wiki tutorials. 🙇

m5o:
Ok, then we stop here. Is the ‘public API for registering custom components’ you mentioned tangible in the near future? Don’t know what exactly to expect here. 

I was following the [Adding custom components](https://github.com/plataformatec/simple_form/wiki/Adding-custom-components) wiki tutorial and happy it worked on the first try. Split the initialier in multiple files don’t make it better.

rafaelfranca:
Not really. Shipping a monkey patch by default frozen us to change the implementation in future releases, so I'm more comfortable with adding a public API. @feliperenan said he would take a look on it

Issue #1176, opened 2014-12-10
Make possible to register custom components.


rafaelfranca:
Right now to register custom components people have to use private API like:

``` ruby
SimpleForm::Inputs::Base.send(:include, MyCustomComponent)
```

I'd like to introduce a public API to make this easier and document this feature.

feliperenan:
Does it still make sense @rafaelfranca ?

feliperenan:
Resolved by https://github.com/plataformatec/simple_form/pull/1556

PR #1556, opened 2018-03-09
Add API to register custom components

feliperenan:
Related to https://github.com/plataformatec/simple_form/issues/1176 and more discussion about it in https://github.com/plataformatec/simple_form/pull/1554

This is a public API to allow users register custom components without monkey patching Simple Form.

### Examples

- Create your custom component.
```ruby
# lib/components/input_group_component.rb
module InputGroup
  def preprend
    ...
  end

  def append
    ...
  end
end

# Register the component in Simple Form.
SimpleForm.include_component(InputGroup)
```

- Tells to Simple Form  the place that the components will be
```ruby
# app/config/initializers/simple_form.rb
Dir[Rails.root.join('lib/components/**/*.rb')].each { |f| require f }
```

- Use the custom component in some wrapper.
```ruby
# app/config/initializers/simple_form.rb
SimpleForm.setup do |config|
  # Create a wrapper using the custom component.
  config.wrappers :input_group, tag: :div, error_class: :error do |b|
    b.use :label
    b.optional :prepend
    b.use :input
    b.use :append
  end
end
```

- Using wrapper with the custom component.

```ruby
<%= simple_form_for @blog, wrapper: input_group do |f| %>
  <%= f.input :title, prepend: true %>
<% end %>
```

### Notes

For while, it is just a idea. I'm open for suggestions 😄 

### Todo
- [x] Add tests 
- [x] Add docs
- [x] Update generator

### DIFF
--- a/README.md
+++ b/README.md
@@ -989,6 +989,92 @@ when the content is present.
   end
 ```
 
+## Custom Components
+
+When you use custom wrappers, you might also be looking for a way to add custom components to your 
+wrapper. The default components are:
+
+```ruby
+:label         # The <label> tag alone
+:input         # The <input> tag alone
+:label_input   # The <label> and the <input> tags
+:hint          # The hint for the input
+:error         # The error for the input
+```
+
+A custom component might be interesting for you if your views look something like this:
+
+```erb
+<%= simple_form_for @blog do |f| %>
+  <div class="row">
+    <div class="span1 number">
+      1
+    </div>
+    <div class="span8">
+      <%= f.input :title %>
+    </div>
+  </div>
+  <div class="row">
+    <div class="span1 number">
+      2
+    </div>
+    <div class="span8">
+      <%= f.input :body, as: :text %>
+    </div>
+  </div>
+<% end %>
+```
+
+A cleaner method to create your views would be:
+
+```erb
+<%= simple_form_for @blog, wrapper: :with_numbers do |f| %>
+  <%= f.input :title, number: 1 %>
+  <%= f.input :body, as: :text, number: 2 %>
+<% end %>
+```
+
+To use the number option on the input, first, tells to Simple Form the place where the components
+will be:
+
+``` ruby
+# config/initializers/simple_form.rb
+Dir[Rails.root.join('lib/components/**/*.rb')].each { |f| require f }
+```
+
+Create a new component within the path specified above:
+
+```ruby
+# lib/components/numbers_component.rb
+module NumbersComponent
+  # To avoid deprecation warning, you need to make the wrapper_options explicit
+  # even when they won't be used.
+  def number(wrapper_options = nil)
+    @number ||= begin
+      options[:number].to_s.html_safe if options[:number].present?
+    end
+  end
+end
+
+SimpleForm.include_component(NumbersComponent)
+```
+
+Finally, add a new wrapper to the config/initializers/simple_form.rb file:
+
+```ruby
+config.wrappers :with_numbers, tag: 'div', class: 'row', error_class: 'error' do |b|
+  b.use :html5
+  b.use :number, wrap_with: { tag: 'div', class: 'span1 number'}
+  b.wrapper tag: 'div', :class: 'span8' do |ba|
+    ba.use :placeholder
+    ba.use :label
+    ba.use :input
+    ba.use :error, wrap_with: { tag: 'span', class: 'help-inline' }
+    ba.use :hint,  wrap_with: { tag: 'p', class: 'help-block' }
+  end
+end
+```
+
 ## HTML 5 Notice
 
 By default, **Simple Form** will generate input field types and attributes that are supported in HTML5,
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -265,6 +265,49 @@ See https://github.com/plataformatec/simple_form/pull/997 for more information.
     @@configured = true
     yield self
   end
+
+  # Includes a component to be used by Simple Form. Methods defined in a
+  # component will be exposed to be used in the wrapper as Simple::Components
+  #
+  # Examples
+  #
+  #    # The application needs to tell where the components will be.
+  #    Dir[Rails.root.join('lib/components/**/*.rb')].each { |f| require f }
+  #
+  #    # Create a custom component in the path specified above.
+  #    # lib/components/input_group_component.rb
+  #    module InputGroupComponent
+  #      def prepend
+  #        ...
+  #      end
+  #
+  #      def append
+  #        ...
+  #      end
+  #    end
+  #
+  #    SimpleForm.setup do |config|
+  #      # Create a wrapper using the custom component.
+  #      config.wrappers :input_group, tag: :div, error_class: :error do |b|
+  #        b.use :label
+  #        b.optional :prepend
+  #        b.use :input
+  #        b.use :append
+  #      end
+  #    end
+  #
+  #    # Using the custom component in the form.
+  #    <%= simple_form_for @blog, wrapper: input_group do |f| %>
+  #      <%= f.input :title, prepend: true %>
+  #    <% end %>
+  #
+  def self.include_component(component)
+    if Module === component
+      SimpleForm::Inputs::Base.include(component)
+    else
+      raise TypeError, "SimpleForm.include_component expects a module but got: #{component.class}"
+    end
+  end
 end
 
 require 'simple_form/railtie' if defined?(Rails)
--- a/test/components/custom_components_test.rb
+++ b/test/components/custom_components_test.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+# Module that represents a custom component.
+module Numbers
+  def number(wrapper_options = nil)
+    @number ||= options[:number].to_s.html_safe
+  end
+end
+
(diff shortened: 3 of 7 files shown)


lib/simple_form.rb
```diff
@@ -265,6 +265,13 @@ def self.setup
     @@configured = true
     yield self
   end
+
+  # TODO include docs
+  def self.register_component(*components)
```
feliperenan: We could include it inside `SimpleForm.setup`, something like:

```ruby
# Register the component in Simple Form.
SimpleForm.setup do |config|
  config.include_component(MyCustomComponent)
end
```

🤔

lib/simple_form.rb
```diff
@@ -265,6 +265,13 @@ def self.setup
...
+    components.each do |component|
```
feliperenan: It doesn't need to be an array, I will change it to receive a single component.

lib/simple_form.rb
```diff
@@ -265,6 +265,49 @@ def self.setup
...
+  # component will be exposed to be used in the wrapper as Simple::Components
+  #
+  # Examples
+  #
+  #    # The application needs to tell where the components will be.
+  #    Dir[Rails.root.join('config/initializers/simple_form_components/**/*.rb')].each { |f| require f }
+  #
+  #    # Create a custom component in the path specified above.
+  #    # config/initializers/simple_form_components/input_group.rb
+  #    module InpoutGroup
```
m5o: ⚡️ typo `InpoutGroup`

lib/simple_form.rb
```diff
@@ -265,6 +265,49 @@ def self.setup
...
     yield self
   end
+
+  # Includes a component to be used by Simple Form. Methods defined in a
+  # component will be exposed to be used in the wrapper as Simple::Components
+  #
+  # Examples
+  #
+  #    # The application needs to tell where the components will be.
+  #    Dir[Rails.root.join('config/initializers/simple_form_components/**/*.rb')].each { |f| require f }
```
m5o: Would it possible to load components from `app/components/**/*.rb` to have feature parity with custom inputs?

lib/simple_form.rb
```diff
@@ -265,6 +265,49 @@ def self.setup
...
+  #
+  #    # Create a custom component in the path specified above.
+  #    # config/initializers/simple_form_components/input_group.rb
```
m5o: to follow rails naming convention the file should be prefix with `_component.rb` => `input_group_component.rb`

lib/simple_form.rb
```diff
@@ -265,6 +265,49 @@ def self.setup
...
+  #      end
+  #
+  #      def append
+  #        ...
+  #      end
+  #    end
+  #    
+  #    SimpleForm.setup do |config|
+  #      # Create a wrapper using the custom component.
+  #      config.wrappers :input_group, tag: :div, error_class: :error do |b|
```
victor-am: Out of curiosity, what does `b` stand for? 🤔

README.md
```diff
@@ -989,6 +989,83 @@ when the content is present.
...
+  end
+end
+
+SimpleForm.include_component(Numbers)
+```
+
+Add a new wrapper to the config/initializers/simple_form.rb file.
+
+```
+config.wrappers :with_numbers, :tag => 'div', :class => 'row', :error_class => 'error' do |b|
```
victor-am: Just a nitpicking, but since this method seems to only configure one wrapper at a time, wdyt of `config.wrapper`? 🤔

lib/simple_form.rb
feliperenan: `b` is the wrapper builder.

You can see more about it here:
https://github.com/plataformatec/simple_form/blob/d32cf5fab7e8d2da7eacd98fa8a00017ca1bf07a/lib/simple_form/wrappers/builder.rb#L4

lib/simple_form.rb
feliperenan: It's just a suggestion, the component can be anywhere in your application. You just need to give the correct path to Simple Form.

README.md
feliperenan: They are different APIs. `config.wrappers` is to define a new wrapper already `config.wrapper` is to retrieve some wrapper.

lib/simple_form.rb
m5o: Ok, how do you like the idea suggest loading the components from `app/components` to have feature parity with custom inputs? 🤓

feliperenan: Ahh I got it now. We already recommend adding custom inputs inside `app/inputs` 😅. This way, I think it makes sense. 👍

README.md
victor-am: Oh, right, you weren't creating the method here, you're just taking use of it 👍

README.md
```diff
@@ -989,6 +989,92 @@ when the content is present.
...
+      <%= f.input :body, :as => :text %>
+    </div>
+  </div>
+<% end %>
+```
+
+A cleaner method to create your views would be:
+
+```erb
+<%= simple_form_for @blog, :wrapper => :with_numbers do |f| %>
```
victor-am: Wdyt of using the new hash syntax here (since the README.md is following it):
`:wrapper => :with_numbers` -> `wrapper: :with_numbers`

feliperenan: I forgot this line 😭

test/components/custom_components_test.rb
```diff
@@ -0,0 +1,62 @@
...
+    assert_select 'div.input-group-append > span.input-group-text'
+  end
+
+  test 'raises a TypeError when the component is not a Module' do
+    component = 'MyComponent'
+
+    exception = assert_raises TypeError do 
+      SimpleForm.include_component(component)
+    end
+    assert_equal exception.message, "SimpleForm.include_component expects a module but got: #{component.inspect}"
```
tegon: WDYT to check the actual message here without interpolating the string?
```
assert_equal exception.message, "SimpleForm.include_component expects a module but got: \"MyComponent\""
```
IMO it makes the test clear since we don't have to follow any logic to discover how the message will look like.

tegon: By the way, I just realized that the exception message looks a little strange 🤔 
This is always going to print the object's name, like the example above. But since we got an object that isn't a module maybe we should print the object's type instead?
For example:
```ruby
❯ component = 'MyComponent'
=> "MyComponent"
❯ "SimpleForm.include_component expects a module but got: #{component.class}"
=> "SimpleForm.include_component expects a module but got: String"
```
WDYT?

README.md
```diff
@@ -989,6 +989,92 @@ when the content is present.
...
+    <div class="span8">
+      <%= f.input :title %>
+    </div>
+  </div>
+  <div class="row">
+    <div class="span1 number">
+      2
+    </div>
+    <div class="span8">
+      <%= f.input :body, :as => :text %>
```
tegon: Old hash syntax 😄

README.md
```diff
@@ -989,6 +989,92 @@ when the content is present.
...
+
+SimpleForm.include_component(NumbersComponent)
+```
+
+Finally, add a new wrapper to the config/initializers/simple_form.rb file:
+
+```ruby
+config.wrappers :with_numbers, tag: 'div', class: 'row', error_class: 'error' do |b|
+  b.use :html5
+  b.use :number, :wrap_with => { tag: 'div', class: 'span1 number'}
```
tegon: Old hash syntax 😄

feliperenan: Finally, the last one!

README.md
```diff
@@ -989,6 +989,92 @@ when the content is present.
...
+  <%= f.input :body, as: :text, number: 2 %>
+<% end %>
+```
+
+To use the number option on the input, first, tells to Simple Form the place where the components 
+will be:
+
+``` ruby
+# config/initializers/simple_form.rb
+Dir[Rails.root.join('app/components/**/*.rb')].each { |f| require f }
```
rafaelfranca: `app/components` will be in a eager load path, do we need to require? Also it will be loaded at boot what may cause autoload problems. Maybe we want to put them in lib to avoid gotchas with autoload?

lib/simple_form.rb
```diff
@@ -265,6 +265,49 @@ def self.setup
...
+  #    end
+  #
+  #    # Using the custom component in the form.
+  #    <%= simple_form_for @blog, wrapper: input_group do |f| %>
+  #      <%= f.input :title, prepend: true %>
+  #    <% end %>
+  #
+  def self.include_component(component)
+    if Module === component
+      SimpleForm::Inputs::Base.send(:include, component)
```
rafaelfranca: I think `include` is a public method now, so we don't need to call `send`

feliperenan: Cool. I didn't know that 😅

README.md
feliperenan: Even inside `app` folder we need to require the file, without that it doesn't work.

I don't know if it will cause autoload problems, do you think it's better put the components inside `lib/components/`?

m5o: @feliperenan If you need a reference implementation, I've granted you access to [m5o/simple_form-bootstrap](https://github.com/m5o/simple_form-bootstrap) this is working example.

rafaelfranca: Yes. `lib` is the correct place for it.

feliperenan: @rafaelfranca Done!. Do you think we are ready to go ahead?

PR #1557, opened 2018-03-21
Add " require 'simple_form' " in template.

sachinswagh:


feliperenan:
Hi @sachinswagh, thanks for the pull-request.

Could you please explain the motivation behind this change?

sachinswagh:
To avoid uninitialized constant error.

sachinswagh:
@feliperenan : If you don't feel any need. You can close the pull request.

feliperenan:
Did you have some use case that you had to require Simple Form?

rafaelfranca:
In a Rails application, the gem is required when you add the `gem` call in the `Gemfile`. That may not be true if people changes the away bundler works in a Rails application, but we don't need to support that case.

PR #1560, opened 2018-03-26
Use custom i18n scope for label required html

tvdeyen:
Since 3.1 it is possible to set a custom `i18n_scope`. This was not
respected to build the required html for label inputs.

### DIFF
--- a/lib/simple_form/components/labels.rb
+++ b/lib/simple_form/components/labels.rb
@@ -7,18 +7,24 @@ module SimpleForm
       module ClassMethods #:nodoc:
         def translate_required_html
           i18n_cache :translate_required_html do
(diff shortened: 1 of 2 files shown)


feliperenan:
@tvdeyen Thanks for the pull request 💪 

What do you think about update the documentation about this feature?. I think you can put it on this section: https://github.com/plataformatec/simple_form#i18n

tvdeyen:
> What do you think about update the documentation about this feature?

@feliperenan I won't consider this PR adding a feature. The feature itself is very old. The `i18n_scope` is a feature introduced in 3.1. It has no mention in the README at all, yet. This PR just "fixes" the missing implementation of the `i18n_scope` feature for the required label html feature.

That said, I don't hink this is in the scope of this PR.

PR #1564, opened 2018-04-03
Allow validation classes on input_field

feliperenan:
This PR is similar to this one https://github.com/plataformatec/simple_form/pull/1552. It allows validation classes on `input_field`.

To get this work I added two configs for the Simple Form: `input_field_error_class` and `input_field_valid_class`. Once the user setup a class for these configs, the configured classes will be rendered according to the form validation.

## Examples

- Setup the `input_field` classes.
```ruby
# config/initializers/simple_form.rb
SimpleForm.setup do |config|
  config.input_field_valid_class = 'is-valid'
  config.input_field_error_class = 'is-invalid'
end
```

- Build the form.
```erb
<%= simple_form_for @product do |f| %>
  <%= f.input_field :name %>

  <div class="form-actions">
    <%= f.button :submit %>
  </div>
<% end %>
```
- When the validation happening, the input will be rendered with the validation classes.
```HTML
<-- Valid example -->
<input class=" is-valid string required" type="text" value="PS4" name="product[name]" id="product_name">

<-- Invalid example -->
<input class=" is-invalid string required" type="text" value="" name="product[name]" id="product_name">
```

## Todo
- [x] Add docs
- [x] Update changeling

### DIFF
--- a/test/form_builder/input_field_test.rb
+++ b/test/form_builder/input_field_test.rb
@@ -173,4 +173,29 @@ class InputFieldTest < ActionView::TestCase
       assert_select 'input[readonly="readonly"]'
     end
   end
+
+  test 'adds valid class to input_field when it is configured' do
+    swap SimpleForm, input_field_valid_class: 'is-valid' do
+      @user.instance_eval { undef errors }
+      with_input_field_for @user, :name
+
+      assert_select 'input.string.required.is-valid'
+    end
+  end
+
+  test 'adds error class to input_field when it is configured' do
+    swap SimpleForm, input_field_error_class: 'is-invalid' do
+      with_input_field_for @user, :name
+
+      assert_select 'input.string.required.is-invalid'
+    end
+  end
+
+  test 'does not add validation classes to input_field when it is not configured' do
+    swap SimpleForm, input_field_error_class: nil, input_field_valid_class: nil do
+      with_input_field_for @user, :name
+
+      assert_select 'input.string.required'
+    end
+  end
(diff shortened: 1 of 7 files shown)


feliperenan:
Hey @remofritzsche. Could you test it?

feliperenan:
@m5o I will appreciate your review as well ❤️

remofritzsche:
It works like a charm, thanks a lot @feliperenan and @m5o for tackling this! We'd love a release with this as soon as possible.

test/form_builder/input_field_test.rb
```diff
@@ -173,4 +173,29 @@ def with_input_field_for(object, *args)
       assert_select 'input[readonly="readonly"]'
     end
   end
+
+  test 'adds valid class to input_field when it is configured' do
+    @user.instance_eval { undef errors }
+    SimpleForm.input_field_valid_class = 'is-valid'
```
tegon: WDYT to use [`swap`](https://github.com/plataformatec/simple_form/blob/dd05b56df12c7a21368f7870b1cff92abc793656/test/support/misc_helpers.rb#L15) here to ensure the state is cleaned after the test runs?
```ruby
swap SimpleForm, input_field_valid_class: 'is-valid' do
  @user.instance_eval { undef errors }
  with_input_field_for @user, :name
  assert_select 'input.string.required.is-valid'
end

feliperenan: Good catch 👍

PR #1582, opened 2018-05-13
Improve disabled option to input_field

### LINKED ISSUE #1365: A single select option cannot be disabled
When creating a select field, I'm trying to disable an option this way:

`f.input_field :foo, label: false, as: :select, collection: Foo.all, disabled: 'bar'`

It is disabling the option with the 'bar' value, but it is also disabling the entire select.

betelgeuse:
Allows individual options to be disabled instead of just the whole
select.

Fixes #1365.

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -165,7 +165,7 @@ module SimpleForm
       components = (wrapper.components.map(&:namespace) & ATTRIBUTE_COMPONENTS)
 
       options = options.dup
-      options[:input_html] = options.except(:as, :boolean_style, :collection, :label_method, :value_method, :prompt, *components)
+      options[:input_html] = options.except(:as, :boolean_style, :collection, :disabled, :label_method, :value_method, :prompt, *components)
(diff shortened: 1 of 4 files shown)


doits:
Is it possible that with this change `disabled: false` is not passed down to the `Tag` as an option?

I have this field:

```rb
f.input_field :group, as: :hidden, disabled: false
```

Before, I could see the key in `ActionView::Helpers::Tags::Base` (see https://github.com/Stellenticket/strong_form/blob/c91f51e50083c2a95088f59ea533fef4bd2fe1fe/lib/strong_form/tag.rb#L19)

Now the key is not present anymore. Can it be that `false` values are filtered out now and not passed down?

doits:
to be more precise: this only happens with `disabled: false`. With `disabled: true` it can be seen there (though with `disabled: 'abc'` it is not there, either 😕)

tegon:
@doits Would you mind opening a new issue for this? It's hard for us to keep track of comments in closed issues and pull requests.

doits:
@tegon sure will do so. Will need to find a simple test case to reproduce before, though.

doits:
done with #1623

PR #1585, opened 2018-05-24
Allow custom error on forms without model

victorperez:
This PR change the `has_errors?` method to allow custom errors on forms without a model:

### Before this PR

```
= simple_form_for :user, url: users_path do |f|
  = f.input :email, error: 'custom error'
```

##### Output

```
...
<div class="form-group email optional user_email">
  <label class="control-label email optional" for="user_email">Email</label>
  <input class="form-control string email optional" type="email" name="user[email]" id="user_email">
</div>
...
```

### After this PR

```
= simple_form_for :user, url: users_path do |f|
  = f.input :email, error: 'custom error'
```

##### Output

```
<div class="form-group email optional user_email has-error">
  <label class="control-label email optional" for="user_email">Email</label>
  <input class="form-control string email optional" aria-invalid="true" type="email" name="user[email]" id="user_email">
  <span class="help-block">custom error</span>
</div>
```

### DIFF
--- a/lib/simple_form/components/errors.rb
+++ b/lib/simple_form/components/errors.rb
@@ -11,7 +11,7 @@ module SimpleForm
       end
 
       def has_errors?
-        object && object.respond_to?(:errors) && errors.present?
+        object_with_errors? || object.nil? && has_custom_error?
       end
 
       def has_value?
@@ -32,6 +32,10 @@ module SimpleForm
 
       def full_error_text
         has_custom_error? ? options[:error] : full_errors.send(error_method)
+      end
+
+      def object_with_errors?
+        object && object.respond_to?(:errors) && errors.present?
       end
 
       def error_method
--- a/test/form_builder/error_test.rb
+++ b/test/form_builder/error_test.rb
@@ -224,6 +224,12 @@ class ErrorTest < ActionView::TestCase
     assert_no_select 'span.error'
   end
 
+  test 'input with custom error works when form does not use a model' do
+    with_form_for :user, :active, error: "Super User Active! cannot be blank"
+
+    assert_select 'span.error'
+  end
+
   test 'input with custom error works when using full_error component' do
     swap_wrapper :default, custom_wrapper_with_full_error do
       error_text = "Super User Name! cannot be blank"


feliperenan:
Hi @victorperez, thanks for the pull request :)

Using the custom error option on a non-model the form ever will be displayed with errors since we don't have validations inside this kind of object, so I'm not sure if it makes sense with a non-model.

Could you explain what use case motivated you to send this PR?

victorperez:
Hi @feliperenan, thanks for the feedback.

Consider the case where you have two applications in your stack, the main application has a view that use `simple_form` without a model and when the form is submitted it will send a request through an API to the second application where the model/validations exists, the second application will create the record or return the errors in case the form is invalid.

If errors are returned from the second app then will be easy to create the errors in the form if we allow custom errors for forms without a model. Of course the idea is to pass a dynamic variable based in the API response to the input (not a hardcoded string that always display the error in the form) ej:

```
= simple_form_for :user, url: users_path do |f|
  ...
  = f.input :first_name, error: errors[:first_name]
  ...
```

If `errors[:first_name]` is `nil` no error will be displayed and if `errors[:first_name]` has a value from the api response then the error will be displayed.

I consider this a weird case and it can be fixed creating a model in the main application to handle the form or using activeresource (https://github.com/rails/activeresource) which is what I finally did in my project, but I consider to open a PR because maybe it could be useful. 

I understand that this is not a common scenario or maybe there is a better approach to handle what I'm exposing so feel free to take the best decision :)

feliperenan:
Thanks for the explanation, now it makes sense to me 👍.

lib/simple_form/components/errors.rb
```diff
@@ -34,6 +34,14 @@ def full_error_text
         has_custom_error? ? options[:error] : full_errors.send(error_method)
       end
 
+      def form_with_object_has_errors?
```
feliperenan: We could name this method of `model_with_errors?`, what do you think?

lib/simple_form/components/errors.rb
```diff
@@ -11,7 +11,7 @@ def full_error(wrapper_options = nil)
       end
 
       def has_errors?
-        object && object.respond_to?(:errors) && errors.present?
+        form_with_object_has_errors? || form_without_object_has_errors?
```
feliperenan: Why do we need to check if the object is nil at the second condition?, just check `has_custom_errors?` would be enough, is it?

lib/simple_form/components/errors.rb
victorperez: Yes I like more that name, but what you think to call it `object_with_errors?` (just to respect the same variable name that we use in the method)

lib/simple_form/components/errors.rb
victorperez: I have to check if the object is `nil` to respect other logic based in forms using an object and if there is not error on the attribute. If the `object.nil?` is removed from the second condition then we have two test failing.

Test 1: https://github.com/victorperez/simple_form/blob/master/test/form_builder/error_test.rb#L221
Test 2: https://github.com/victorperez/simple_form/blob/master/test/form_builder/error_test.rb#L274

feliperenan: So we could rename this method or just use the conditions here, something like this:

```ruby
object_with_errors? || object.nil? && has_custom_error?
```

victorperez: Done, I have amended my commit changed the conditions to what you propose 👍

lib/simple_form/components/errors.rb
victorperez: Done, I have amended my commit and renamed the method.

PR #1590, opened 2018-06-07
Fix column type citext HTML5 input type bug

### LINKED ISSUE #1589: citext column type does not set HTML5 input type properly
## Environment

- Ruby 2.4.2
- Rails 5.1.1
- Simple Form 4.01

## Current behavior
A database column of type :citext renders the following html5 input (User.first_name example):
`<input class="string citext required" type="citext" value="" name="user[first_name]" id="user_first_name" data-validate="true">`
Note that the input type is citext, which I don't think is a valid HTML5 input type and in my case causes Foundation to fail to apply the proper CSS to the input.

## Expected behavior
Expected input example:
`<input class="string citext required" type="text" value="" name="user[first_name]" id="user_first_name" data-validate="true">`

## Potential Fix
Updating the string? method in SimpleForm::Inputs::StringInput seems to be the easiest fix:
```     
def string?
  input_type == :string || input_type == :citext
end
```

brucew:
Fixes #1589

### DIFF
--- a/lib/simple_form/inputs/string_input.rb
+++ b/lib/simple_form/inputs/string_input.rb
@@ -18,7 +18,7 @@ module SimpleForm
       private
 
       def string?
-        input_type == :string
+        input_type == :string || input_type == :citext
       end
     end
   end


feliperenan:
I'm good with your solution. With that we can remove the `citext` of the [this condition](https://github.com/swrobel/simple_form/blob/1ca0fb46afd4b3398f73573e4eeb80fb45276f7f/lib/simple_form/form_builder.rb#L530).

Would be nice if we cover this change with unit tests. Maybe asserting the type in [this spec](https://github.com/plataformatec/simple_form/pull/1530/files#diff-c230aabcf94303a1109922d8482c8b24R187) would be enough 🤔.

**PS:** You can use the keyword `Fixes #1589` in the PR description to close the issue automatically when this PR gets merged

brucew:
> With that we can remove the citext of this condition. Could you do that?
This case statement is based off of column.type, which should still be :citext so it should be left alone or citext email columns will not use an input type of email, for example.

feliperenan:
@brucew Yeah, that's true. So, just cover the change with unit test and we can go ahead  :)

PR #1599, opened 2018-08-19
Allows config.wrapper_mappings to allow wrapper: false

dgilperez:
Given a custom SimpleForm input `class MyCustomInput < SimpleForm::Inputs::Base; end`

It appears a `wrapper_mappings` can't be configured with the option of `wrapper: false`. 

Passing the option to `input`, or using `input_field` will work

```
<%= f.input :attribute, as: :my_custom, wrapper: false %>
<%= f.input_field :attribute %>
```

but setting `wrapper_mappings` to disable the wrapper, like this

```
# config/initializers/simple_form.rb
config.wrapper_mappings = {
  my_custom: false
}
```

will not work: wrapper will be still rendered. This behaviour feels unexpected and not consistent with `f.input`. 

This PR allows for this setting to be set. 

If behaviour described was intended for any reason, and PR is not to be accepted, or if there is something I can improve in the PR, please let me know. 

_Originated by this question at SO:_ https://stackoverflow.com/questions/51916852/how-to-configure-a-simpleform-input-to-default-to-no-wrapper/51917096#51917096

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -616,7 +616,7 @@ module SimpleForm
     # 1) It tries to find a wrapper for the current form
     # 2) If not, it tries to find a config
     def find_wrapper_mapping(input_type)
-      if options[:wrapper_mappings] && options[:wrapper_mappings][input_type]
+      if options[:wrapper_mappings] && !options[:wrapper_mappings][input_type].nil?
         options[:wrapper_mappings][input_type]
       else
         SimpleForm.wrapper_mappings && SimpleForm.wrapper_mappings[input_type]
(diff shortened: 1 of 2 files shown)


dgilperez:
Anything I can do here to move this PR? Something's missing?

feliperenan:
Sorry for the delay @dgilperez. I will try to review this PR this week.

feliperenan:
Hi @dgilperez 

I don't think that we should add it to Simple Form since we have different ways to disable the wrapper when we need to.

If the developer wants to a custom wrapper by default, he/she can create one and changes the default as you mentioned in the Stack Overflow. 

Thank for the pull-request anyway :).

dgilperez:
Ok, understood. I still think current behaviour feels unexpected and this addition would make the API more homogeneous ... saving some of other people's time in the process. But it's also kind of an edge thing, maybe not worth the extra code. Thanks for reviewing anyway!

Issue #1604, opened 2018-09-18
input_html_classes called multiple times in custom inputs leading to duplicate classes


RigoTheDev:
## Environment

- Ruby: **ruby 2.5.1p57**
- Rails: **5.2.1**
- Simple Form: **4.0.1**

## Current behavior

I'm trying to add an extra class to `TextInput` by overriding `SimpleForm::Inputs::TextInput` like so

```ruby
class TextInput < SimpleForm::Inputs::TextInput
  def input_html_classes
    super.push('textarea')
  end
end
```

It works, but the class is being added to the `textarea` element 3 times, instead of once.

```html
<textarea class="text optional textarea textarea textarea"></textarea>
```

## Expected behavior

It should only include the custom class once:

```html
<textarea class="text optional textarea"></textarea>
```

RigoTheDev:
Only happens if `config.input_class` is not `nil`.

RigoTheDev:
The reason this only happens if `config.input_class` is not `nil` is because this block of code checks for `SimpleForm.input_class` before evaluating `input_html_classes` again twice. Note that the right side of the expression: `!input_html_classes.empty?` will never be evaluated if the left side returns `false` due to short-circuit evaluation.

https://github.com/plataformatec/simple_form/blob/1411dc3349ea03d4f3e902b9340187e4133067e8/lib/simple_form/inputs/base.rb#L75-L77

PR #1605, opened 2018-09-18
Memoize input_html_classes

RigoTheDev:
Fixes: #1604. `input_html_classes` is a function therefore calling it multiple times will run the function multiple times.

This pull request stores the return value of `input_html_classes` in a variable so that it is called only once. That variable can then be used instead.

### DIFF
--- a/lib/simple_form/inputs/base.rb
+++ b/lib/simple_form/inputs/base.rb
@@ -73,13 +73,13 @@ module SimpleForm
 
         @input_html_classes = @html_classes.dup
 
-        input_html_classes_memoized = input_html_classes
+        input_html_classes = self.input_html_classes
 
-        if SimpleForm.input_class && !input_html_classes_memoized.empty?
-          input_html_classes_memoized << SimpleForm.input_class
+        if SimpleForm.input_class && input_html_classes.any?
+          input_html_classes << SimpleForm.input_class
         end
 
-        @input_html_options = html_options_for(:input, input_html_classes_memoized).tap do |o|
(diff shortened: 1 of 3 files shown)


lib/simple_form/inputs/base.rb
```diff
@@ -72,11 +72,14 @@ def initialize(builder, attribute_name, column, input_type, options = {})
         @html_classes = SimpleForm.additional_classes_for(:input) { additional_classes }
 
         @input_html_classes = @html_classes.dup
-        if SimpleForm.input_class && !input_html_classes.empty?
-          input_html_classes << SimpleForm.input_class
+
+        input_html_classes_memoized = input_html_classes
```
rafaelfranca: can we just call this just `input_html_classes = self.input_html_classes`?

lib/simple_form/inputs/base.rb
```diff
@@ -72,11 +72,14 @@ def initialize(builder, attribute_name, column, input_type, options = {})
...
+
+        if SimpleForm.input_class && !input_html_classes_memoized.empty?
```
feliperenan: Since we are here, we could use `input_html_classes.any?` rather than `!input_html_classes.empty?`.

test/inputs/discovery_test.rb
```diff
@@ -109,6 +109,18 @@ def discovery(value = false)
     end
   end
 
+  test 'giving extra class to input_html_classes adds it to element only once' do
```
feliperenan: Wdty of: `does not duplicate the html classes giving a extra class` 🤔 ?

PR #1606, opened 2018-09-18
Create setting to be able to remove input_type classes from generated elements

RigoTheDev:
This configuration setting allows you to remove input_type classes from generated wrappers, labels, & inputs. All existing functionality stays the same so this change is non-breaking.

By default, all generated inputs look like this:

```html
<div class="field text optional item_description">
  <label class="text optional" for="item_description">Description</label>
  <textarea class="text optional" name="item[description]" id="item_description"></textarea>
</div>
```

By adding this configuration in `config/initializers/simple_form.rb`:
```ruby
  ...
  config.generate_input_type_class = false # Note that by default it's true.
  ...
```

...all generated inputs now look like this:
```html
<div class="field optional item_description">
  <label class="optional" for="item_description">Description</label>
  <textarea class="optional" name="item[description]" id="item_description"></textarea>
</div>
```

Also note that `optional/required` classes stay the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Enhancements
 * Guess input type more carefully. [sringling](https://github.com/sringling)
 * Allow custom error on forms without model. [victorperez](https://github.com/victorperez)
+* Allow removing input_type classes from generated elements. [@RigoTheDev](https://github.com/RigoTheDev)
 
 ### Bug fix
 * Improve disabled option to input_field. [betelgeuse](https://github.com/betelgeuse)
--- a/lib/generators/simple_form/templates/config/initializers/simple_form.rb
+++ b/lib/generators/simple_form/templates/config/initializers/simple_form.rb
@@ -119,6 +119,10 @@ SimpleForm.setup do |config|
   # You can define which elements should obtain additional classes
   # config.generate_additional_classes_for = [:wrapper, :label, :input]
 
+  # You can define whether generated elements contain input type classes.
+  # Default is true.
+  # config.generate_input_type_class = true
+
   # Whether attributes are required by default (or not). Default is true.
   # config.required_by_default = true
 
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -112,6 +112,10 @@ See https://github.com/plataformatec/simple_form/pull/997 for more information.
   mattr_accessor :generate_additional_classes_for
   @@generate_additional_classes_for = %i[wrapper label input]
 
+  # You can define whether generated inputs contain input type classes
+  mattr_accessor :generate_input_type_class
+  @@generate_input_type_class = true
+
   # Whether attributes are required by default or not.
   mattr_accessor :required_by_default
   @@required_by_default = true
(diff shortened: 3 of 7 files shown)


rafaelfranca:
Thank you for the pull request but we want to avoid adding more configuration options to this gem, unless there is a good reason. Not generating a class just because you want a clean HTML is not a good reason to add a new option.

RigoTheDev:
@feliperenan Is there a workaround we can use here?

quoyi:
> Thank you for the pull request but we want to avoid adding more configuration options to this gem, unless there is a good reason. Not generating a class just because you want a clean HTML is not a good reason to add a new option.

You want reason: bulma select input wrapped by "select"

quoyi (commented):
THX

hidde-jan:
Just so the maintainers know what is needed to work around this issue. I've added the following monkeypatch to my simple_form initializer:

```ruby
module SimpleForm
  module Wrappers
    class Root
      private

      def html_classes(input, options)
        css = options[:wrapper_class] ? Array(options[:wrapper_class]) : @defaults[:class]
        css += SimpleForm.additional_classes_for(:wrapper) do
          classes = input.additional_classes + [input.input_class]
          classes.map { |c| c == input.input_type ? "wrapper-#{c}" : c }
        end
        css << html_class(:error_class, options) { input.has_errors? }
        css << html_class(:hint_class, options) { input.has_hint? }
        css << html_class(:valid_class, options) { input.valid? }
        css.compact
      end
    end
  end

  module Components
    module Labels
      module ClassMethods
        def label_html_options
          label_html_classes = SimpleForm.additional_classes_for(:label) {
            [input_type && "label-#{input_type}", required_class, disabled_class, SimpleForm.label_class].compact
          }

          label_options = html_options_for(:label, label_html_classes)
          if options.key?(:input_html) && options[:input_html].key?(:id)
            label_options[:for] = options[:input_html][:id]
          end

          label_options
        end
      end
    end
  end
end

**Comment by hidde-jan** [comment]
The following monkeypatch needs to be added to your initializer to make simple_form work with bulma selects. I hope the maintainers will reconsider this PR.

```ruby
module SimpleForm
  module Wrappers
    class Root
      private

      def html_classes(input, options)
        css = options[:wrapper_class] ? Array(options[:wrapper_class]) : @defaults[:class]
        css += SimpleForm.additional_classes_for(:wrapper) do
          classes = input.additional_classes + [input.input_class]
          classes.map { |c| c == input.input_type ? "wrapper-#{c}" : c }
        end
        css << html_class(:error_class, options) { input.has_errors? }
        css << html_class(:hint_class, options) { input.has_hint? }
        css << html_class(:valid_class, options) { input.valid? }
        css.compact
      end
    end
  end

  module Components
    module Labels
      module ClassMethods
        def label_html_options
          label_html_classes = SimpleForm.additional_classes_for(:label) {
            [input_type && "label-#{input_type}", required_class, disabled_class, SimpleForm.label_class].compact
          }

          label_options = html_options_for(:label, label_html_classes)
          if options.key?(:input_html) && options[:input_html].key?(:id)
            label_options[:for] = options[:input_html][:id]
          end

          label_options
        end
      end
    end
  end
end
```

PR #1613, opened 2018-10-30
Add color input

gssbzn:
Add an input for color picker

Browser support for color picker is still limited but I don't see a reason not to support it within simple_form


lib/simple_form/inputs/color_input.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module SimpleForm
+  module Inputs
+    class ColorInput < Base
+      def input(wrapper_options = nil)
+        input_html_options[:type] ||= "color" if html5?
```
feliperenan: Maybe we don't need to check the `html5?` and always generate the input with the `type="color"`. The responsibility of using it or not would stay with the developer.

feliperenan:
Is this API stable? I mean, is there any change to `type="color"` change? if so, I would avoid it in Simple Form at this moment.

gssbzn:
@feliperenan the API is stable but support is on part with `type="date"`, meaning no IE nor Safari

lib/simple_form/inputs/color_input.rb
gssbzn: I was following the same patterns as [range](https://github.com/plataformatec/simple_form/blob/master/lib/simple_form/inputs/range_input.rb) or [numeric](https://github.com/plataformatec/simple_form/blob/master/lib/simple_form/inputs/numeric_input.rb) but if we don't mind for color then I can change it

feliperenan: So let's keep as it is :)

test/inputs/color_input_test.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+# encoding: UTF-8
```
feliperenan: We don't need this magic comment since we support ruby 2.3 +

gssbzn:
@feliperenan rebased and addressed the UTF magic comment

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Guess input type more carefully. [@sringling](https://github.com/sringling)
 * Allow custom error on forms without model. [@victorperez](https://github.com/victorperez)
 * Do not support Ruby < 2.3 anymore. [@gssbzn](https://github.com/gssbzn)
+* Add color inout type. [@gssbzn](https://github.com/gssbzn)
```
m5o: ⚡️ typo? `inout` 🤔

mustmodify:
I noticed 'color' isn't on this list. Should I add it? https://github.com/plataformatec/simple_form#available-input-types-and-defaults-for-each-column-type

mracos:
@mustmodify please do :)

PR #1625, opened 2018-12-12
Fix example for Collection Radio Buttons

akshatkedia:


nashby:
@akshatkedia thank you! 🥇

PR #1626, opened 2018-12-17
set multiple attribute for grouped selects also

ollym:


feliperenan:
Hi @ollym, thanks for the pull request.

Could add the motivation of this PR? It would be nice to know the use case or if it's a bug. Also, don't forget to cover this change with tests :).

Thanks

ollym:
@felipecbarbosa 

```
= f.association :tags, as: :grouped_select
```

If :tags is a `has_many` or `has_and_belongs_to_many` association then it should automatically add a `multiple` attribute to the select tag, this would make it the same as `as: :select`

Issue #1579, opened 2018-05-10
No input found for money


frankyston:
## Environment

- Ruby **ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]**
- Rails **Rails 5.2.0**
- Simple Form **simple_form (4.0.0)**

## Current behavior

I'm using a template called a lot and there are some money type fields
```
class CreateLots < ActiveRecord::Migration[5.2]
  def change
    create_table :lots do |t|
      t.integer :number_lot
      t.string :title
      t.money :value
      t.money :value_first_squere
      t.money :value_second_square
      t.money :value_increment
      t.integer :modality, default: 0
      t.integer :status, default: 0

      t.timestamps
    end
  end
end
```

when I render the form is giving the following error: `No input found for money`

```
= f.input :value
```

Were you expecting the simple form rendering normally made up, or am I doing something wrong?

feliperenan:
Hi @frankyston, thank you for the report.

Could you please provide us a sample application that reproduces the issue in isolation? 

That would help us find the issue.

benoitr:
Get the same issue

@feliperenan Here a sample app :) :

https://github.com/benoitr/money_app

benoitr:
Just get it (needed to specify the input type)

Reference : https://stackoverflow.com/questions/36847428/simple-form-im-getting-no-input-found-for-citext

    = f.input :price, as: :integer

alexanderadam:
@feliperenan can this issue be reopened now, that the sample app is available sinceJjune of last year?

feliperenan:
We decided to not add this support at Simple Form since money field isn't common. You can create a custom input case you still need it.

See more about how to create custom components here:
https://github.com/plataformatec/simple_form#custom-inputs

Thanks for reporting the issue :)

PR #1630, opened 2019-01-21
Add support to Money field

feliperenan:
Fix https://github.com/plataformatec/simple_form/issues/1579

Since Rails has support to `money` field, I'm adding to Simple Form render this input properly.

### DIFF
--- a/lib/simple_form/inputs/money_input.rb
+++ b/lib/simple_form/inputs/money_input.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+module SimpleForm
+  module Inputs
+    class MoneyInput < Base
+      enable :placeholder, :maxlength, :minlength
+
+      def input(wrapper_options = nil)
+        merged_input_options = merge_wrapper_options(input_html_options, wrapper_options)
+
+        @builder.text_field(attribute_name, merged_input_options)
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


feliperenan:
@frankyston @benoitr @alexanderadam Could some of you test this branch to see if I'm forgetting something? 🙏

lib/simple_form/inputs/money_input.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module SimpleForm
+  module Inputs
+    class MoneyInput < Base
```
feliperenan: Another approach would be handling it inside of `StringInput`. But I created a new input to make it explicit. Does it make sense? 🤔

rafaelfranca:
Since creating a custom input is not hard and I don't think money columns are common enough to require support in the library for them.

feliperenan:
@rafaelfranca Indeed, creating a custom input is pretty straightforward, I didn't think about it 🙈. Do you think we can close this PR then? I'm ok in doing that.

rafaelfranca:
Since the code is here I'm fine with it, but for future it would be better to see how often the type is used to make a decision.

feliperenan:
Since it's not common, I'm going to close this one. If people ask for it we can reopen in the future.

Issue #141, opened 2010-12-09
w3c validation errors due to 'required' attribute on input type select


takaltoo:
The w3c validator (http://validator.w3.org/#validate_by_input) raises the following error on a form generated using simple_form:

Line 15, Column 123: Attribute required not allowed on element select at this point.

The form snippet causing the issue is given below:

<select class="select required" id="photo_display_size"   name="photo[display_size]" required="required">
 <option value="square">square</option>
 <option value="thumbnail">thumbnail</option>
</select>

According to: (http://www.w3schools.com/html5/html5_form_attributes.asp)  the required attribute only works with the following <input> types:  text, search, url, telephone, email, password, date pickers, number, checkbox, radio, and file

Using the "required" attribute on the select tags raises a w3c error.  

The simple_form code used to generate the display_size:

f.input :display_size, :collection => @photo.possible_sizes, :prompt => 'Please select size'

PR #1649, opened 2019-03-28
Allow "required" attribute in generated select elements of PriorityInput

### LINKED ISSUE #340: Ability to make associations and collections required
Setting associations and collections to be required doesn't seem to have any effect on validation at the moment. A default blank option is still displayed, selectable and saveable.

### LINKED ISSUE #866: Select required attribute missing
The `required` attribute is missing from the select box. I tried using 3.0.0.rc.

Given the select box:

``` haml
= f.input :home_country, collection: Country.all, required: true
```

I get:

``` html
<select class="select required" id="account_home_country" name="account[home_country]"></select>
```

I want:

``` html
<select required="required" class="select required" id="account_home_country" name="account[home_country]"></select>
```

I can get around it with:

``` haml
= f.input :home_country, collection: Country.all, input_html: { required: 'required' }
```

### LINKED ISSUE #1355: Simple Form Country Select always optional?
I am using Foundation's abide validation for a form, which looks for the presence of the required attribute. However it appears that the country input doesn't allow you to set the attribute. Is there a particular reason for this? It's inconvenient as a number of validation plugins/frameworks utilise that attribute. 

I'd prefer to use the country select than building my own with countries, so the priority option can be set.

mcountis:
Don't overwrite `#has_required?` in `PriorityInput`, let that input inherit behavior.

Properly close #1355.
Properly close #866.
Continuation of #373 that closed #340.
Undoes some of, very old, outdated, #141.

### DIFF
--- a/lib/simple_form/inputs/priority_input.rb
+++ b/lib/simple_form/inputs/priority_input.rb
@@ -15,10 +15,6 @@ module SimpleForm
 
       protected
 
-      def has_required?
-        false
-      end
-
       def skip_include_blank?
         super || input_priority.present?
       end
--- a/test/inputs/priority_input_test.rb
+++ b/test/inputs/priority_input_test.rb
@@ -36,15 +36,15 @@ class PriorityInputTest < ActionView::TestCase
     assert_no_select 'select option[value=""]', /^$/
   end
 
-  test 'priority input does not generate invalid required html attribute' do
+  test 'priority input does generate select element with required html attribute' do
     with_input_for @user, :country, :country
     assert_select 'select.required'
-    assert_no_select 'select[required]'
+    assert_select 'select[required]'
   end
 
-  test 'priority input does not generate invalid aria-required html attribute' do
+  test 'priority input does generate select element with aria-required html attribute' do
     with_input_for @user, :country, :country
     assert_select 'select.required'
-    assert_no_select 'select[aria-required]'
+    assert_select 'select[aria-required]'
   end
 end


feliperenan:
Hi @mcountis, thanks for the pull-request.

I didn't know that we couldn't use `required` attribute in the select tag in the past. So, this PR seems resonable to me. But It seems that we have some rules when using this attribute as pointed out in this [blog post](http://maxdesign.com.au/articles/select-required/). I'm not sure if we need to worry about them though 🤔 . Have you considered it?

mcountis:
> I didn't know that we couldn't use `required` attribute in the select tag in the past. So, this PR seems reasonable to me. But It seems that we have some rules when using this attribute as pointed out in this [blog post](http://maxdesign.com.au/articles/select-required/). I'm not sure if we need to worry about them though 🤔. Have you considered it?

The resulting `PriorityInput#has_required?` method is defined by `CollectionInput#has_required?`:

https://github.com/plataformatec/simple_form/blob/de09cbf63b7b20f3ff4337aa41e09263ef43c978/lib/simple_form/inputs/collection_input.rb#L43-L45

Which I believe is already adhering to the HTML spec. Where either the `:include_blank` or the `:prompt` options will include a child `<option>` tag with an empty value, or if the `:multiple` option is supplied the constraints on any child `<option>` tags are not considered.

feliperenan:
Yeah, that makes sense. Let's merge it then.

Thanks ❤️ 💙 💚

PR #1651, opened 2019-04-05
Removing default success validation from bootstrap API wrapper.

tunnes:
## Motivation 
In order to solve the premature validation in inputs and wrappers rendering, was solicited by @feliperenan and this issue [1596](https://github.com/plataformatec/simple_form/issues/1596) to remove default success validations from API initializers, now only errors are validated on rendering cause error situations the form is rendering again instead success usually no.

#### Before (Current behavior)
after

#### After (With pull request changes)
before
 

## Goals
- [x] Comment and remove success validation classes from bootstrap initializers.
- [x] Involve signaling for customization.

Issue reference: [1596](https://github.com/plataformatec/simple_form/issues/1596)

### DIFF
--- a/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
+++ b/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
@@ -41,15 +41,16 @@ SimpleForm.setup do |config|
   # :to_sentence to list all errors for each field.
   config.error_method = :to_sentence
 
+  # uncomment the assignment below to add `is-valid` class to valid `input_field`
+  # config.input_field_valid_class = 'is-valid'
+
   # add validation classes to `input_field`
   config.input_field_error_class = 'is-invalid'
-  config.input_field_valid_class = 'is-valid'
 
-
-  # vertical forms
   #
   # vertical default_wrapper
-  config.wrappers :vertical_form, tag: 'div', class: 'form-group', error_class: 'form-group-invalid', valid_class: 'form-group-valid' do |b|
+  # add `valid_class: 'form-group-valid'` below to add `form-group-valid` class into wrapper
+  config.wrappers :vertical_form, tag: 'div', class: 'form-group', error_class: 'form-group-invalid' do |b|
     b.use :html5
     b.use :placeholder
     b.optional :maxlength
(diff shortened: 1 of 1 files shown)


feliperenan:
I'm not sure if we should continue with this solution. If we do, we also need to remove the `valid_class` option from the foundation initializer.

https://github.com/plataformatec/simple_form/blob/14587b516116e5fb3f9d0101e2330adb9e4421b7/lib/generators/simple_form/templates/config/initializers/simple_form_foundation.rb#L18

Given that developers can remove this option from the initializer, I think we could live with this caveat since it's already implemented and released. But if others think this could be disabled by default, we can reconsider in the future.

As it requires changes in the initializer, I think this comment doesn't make sense anymore.

https://github.com/plataformatec/simple_form/blob/14587b516116e5fb3f9d0101e2330adb9e4421b7/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb#L3

IMO, our initializers and [simple_form-bootstrap](# https://github.com/rafaelfranca/simple_form-bootstrap) should be used as a reference and/or a bootstrap since they can change it according to their application needs. That said, I don't think we should incentivize developers to not change their initializer.

What do you think?

cc @m5o

m5o:
> IMO, our initializers and/or [simple_form-bootstrap](https://github.com/rafaelfranca/simple_form-bootstrap) should be used as a reference since developers may change it according to their application needs. That said, I don't think we should incentivize developers to not change their initializer.

This is a strong quote 👍 

@tunnes if you're cooking your application this way, change the bootstrap initializer the way it fit's your application. Totally legit and fine. The requested changes however just comment the `is-valid` css class from the fields. Next one will have the `error_class` optional and so… Not a way we could proceed, IMO.

Of course, a possible topic could be to improve the visibility of those validation classes on `edit` forms when rendered. Keep in mind, inline form feedback might always involve more fine tuning and possible JS behavior to work super smooth. Maybe I'll do some stimulus example in the example app, in the future, no promise.

Remember, if you have a more advanced form, with multiple inputs, it's valuable to have these visual `valid`/`invalid` reference on the fields already on `create`. To spot the ones with errors easier.

Screenshot 2019-04-29 at 15 30 16

feliperenan:
@mracos @tegon Given the last comments, we could close this PR, the issue and remove [this comment](https://github.com/plataformatec/simple_form/blob/14587b516116e5fb3f9d0101e2330adb9e4421b7/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb#L3) from the initializer 😉

Issue #1656, opened 2019-04-23
Skip valid_class check if no class defined


TALlama:
## Environment

- Ruby 2.6.2
- Rails 5.1.7
- Simple Form 4.1.0

## Current behavior

We have a simple file input in our form:

```rb
<%= form.input :content, as: :file %>
```

As of 4.1, this is now a performance problem, because the file in question is stored in S3, and each time we render this form we now fetch the file from S3 to check if it's present, because `simple_form` wants to add the `valid_class`. In our case this is one of several such forms on the same page, which is now prohibitively slow.

We have defined `valid_class: nil` in our wrapper config, but `simple_form` still does the work to determine validity even if no class will be added if we are valid. We also marked the field as `required: false` in hopes that it being optional would allow it to be valid when it's not present, but that doesn't affect anything either.

This is very similar to #1596, except that only talks about getting rid of the CSS class, whereas I want to go a step further and avoid doing the (sometimes expensive) work that determines if a CSS class is necessary.

## Expected behavior

If `valid_class` is `nil` (in the wrapper config or in the input options), no validity check should be performed.

marcelolx:
@TALlama I think this issue can be closed, right?

feliperenan:
Oh yeah, it seems closed. Thanks :)

PR #1657, opened 2019-04-23
Skip valid_class check if no class defined

TALlama:
https://github.com/plataformatec/simple_form/issues/1656

Previously, we would check for validity all the time, even if the outcome of the check added no class. This meant that there was no way to opt out of the check in cases where it was expensive (if the file is stored in S3, in our example).

Now, we only check for validity if there is a valid_class defined to be added if the field is valid.

### DIFF
--- a/test/form_builder/wrapper_test.rb
+++ b/test/form_builder/wrapper_test.rb
@@ -59,6 +59,15 @@ class WrapperTest < ActionView::TestCase
     assert_no_select 'input.is-invalid'
   end
 
+  test 'wrapper does not determine if valid class is needed when it is set to nil' do
+    def @user.name; raise BOOM; end
+    with_form_for @user, :name, as: :file, wrapper: custom_wrapper_with_input_valid_class(valid_class: nil)
+    assert_select 'div'
+    assert_select 'input'
+    assert_no_select 'div.field_with_errors'
+    assert_no_select 'input.is-invalid'
+  end
+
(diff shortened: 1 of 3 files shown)


TALlama:
Anything I can do to expedite review? This issue is currently blocking us from updating `simple_form`, which is blocking us from updating to Rails 5.2.

test/form_builder/wrapper_test.rb
```diff
@@ -59,6 +59,15 @@ class WrapperTest < ActionView::TestCase
     assert_no_select 'input.is-invalid'
   end
 
+  test 'wrapper does not determine if valid class is needed when it is set to nil' do
```
feliperenan: This test is not covering your changes - I put back the deleted code and it still pass. One way I can think right in how to test it, would be with something like:

```ruby
  test 'wrapper does not determine if valid class is needed when it is set to nil' do
    @user.instance_eval { undef errors }
    with_form_for @user, :name, wrapper: custom_wrapper_with_input_valid_class(valid_class: nil)

    assert_no_select 'div.field_without_errors'
  end
```
A better test would checking if `input.valid?` has not been called at all, but I don't know how to do it :see_no_evil:

feliperenan:
Hi @TALlama 

Thanks for your contribution and sorry about the delay :raised_hands: . Before we decide whether or not to move forward with this, I'd like to know if you are still facing this issue or did you work around it somehow?

PR #1663, opened 2019-06-23
Bootstrap uses nested labels for check/radio

jrz:
Using :inline will leave a bottom margin on the label, resulting in a misaligned label for checkboxes / radiobuttons.


m5o:
Would you like to submit the PR at [rafaelfranca/simple_form-bootstrap](https://github.com/rafaelfranca/simple_form-bootstrap) were the generator is maintained. 
Here you see the direct implication of your change within the different form types of bootstrap.

feliperenan:
Hi @jrz

I think that's fine if we keep it the way it is since your can change this config in your own application :raised_hands: 

Thanks for the PR anyway :wave:

Issue #1671, opened 2019-09-30
Inputs are accidentally converted to file inputs after upgrade to 5.0


fschwahn:
## Environment

- Ruby 2.6
- Rails 5.2
- Simple Form 5.0

## Current behavior

When updating to simple_form 5.0, some of our inputs were accidentally converted to file inputs. This was caused by the new check for file fields, namely the check for `#{attribute_name}_url` (eg. a field `twitter_account`, with an accompanying `twitter_account_url`-method). I can imagine that having a `_url` suffix for attributes is not _that uncommon_, so this could bite more people.

I fixed the issue by explictly setting `as: :string` for these inputs.

## Expected behavior

Maybe it would be possible to replace this check with a more specific one, eg. `remote_#{attribute_name}_url=` - I think that would make accidental false positive much less likely. This check would work for both `simple_form` and `refile`.

tegon:
Hi @fschwahn, thanks for reporting this.

Unfortunately, there will always be the possibility of false positives but it's definitely better if we can mitigate it.

I'll try to find out in which versions of the Gems `remote_#{attribute_name}_url` is present to see if it's viable.

Redapted:
Ran into this problem as well. Maybe a config option could be added allowing folks to specify the gem?

For example:
```ruby
SimpleForm.file_attachment_gem = :paperclip
```

That would allow us to use the new, safer strategy, but skip irrelevant checks. Something along the lines of this may work:

```ruby
    ATTACHMENT_GEM_METHODS = {
      active_storage: ['%s_attachment'],
      refile:         ['%s_attachment', '%s_url', '%s_attacher'],
      shrine:         ['%s_url', '%s_attacher'],
      carrier_wave:   ['%s_url'],
      paperclip:      ['%s_file_name']
    }.freeze

    # ...

    def file_method?(attribute_name)
      possible_methods =
        ATTACHMENT_GEM_METHODS.fetch(
          SimpleForm.file_attachment_gem,
          ATTACHMENT_GEM_METHODS.values.flatten.uniq
        )

      possible_methods.any? do |method|
        formatted_method = format(method, attribute_name)
        @object.respond_to?(formatted_method)
      end
    end
```
This is just a rough illustration; it may or may not actually run as-intended (IIRC, `#values` ordering isn't guaranteed to be stable).

tegon:
@Redapted this is more flexible but I'm not sure if the complexity is worth it. 
[...]

PR #1674, opened 2019-10-03
Replace `_url` with `remote_url`

tegon:
As reported in https://github.com/plataformatec/simple_form/issues/1671, the method with suffix `_url` is likely to cause false positives. By using the alternative `remote_url` we might be able to make wrong assumptions, although they're still possible.

Fixes https://github.com/plataformatec/simple_form/issues/1671

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -580,7 +580,7 @@ module SimpleForm
     #
     # Note: This does not support multiple file upload inputs, as this is very application-specific.
     #
-    # The order here was choosen based on the popularity of Gems:
+    # The order here was chosen based on the popularity of Gems:
     #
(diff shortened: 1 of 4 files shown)


lib/simple_form/form_builder.rb
```diff
@@ -580,18 +580,17 @@ def find_custom_type(attribute_name)
     #
     # Note: This does not support multiple file upload inputs, as this is very application-specific.
     #
-    # The order here was choosen based on the popularity of Gems and for commodity - e.g. the method
-    # with the suffix `_url` is present in three Gems, so it's checked with priority:
+    # The order here was choosen based on the popularity of Gems:
```
mracos:
```suggestion
    # The order here was chosen based on the popularity of Gems:
```

tegon:
@rafaelfranca Yeah, I'll start to working on the active storage implementation soon.

PR #1680, opened 2019-12-09
Allow additional form classes

KevinBongart:
This lets you add one or more classes to the form without replacing the default classes. Particularly useful for libraries that require you to add a class to the form, such as [Dropzone.js](https://www.dropzonejs.com/#installation).

```erb
<%= simple_form_for(@user, html: { classes: 'dropzone' }) do |f| %>
   ...
<% end %>
```

⬇️

```html
<form class="simple_form new_user dropzone" ...>
```

### DIFF
--- a/lib/simple_form/action_view_extensions/form_helper.rb
+++ b/lib/simple_form/action_view_extensions/form_helper.rb
@@ -17,11 +17,14 @@ module SimpleForm
         unless options[:html].key?(:novalidate)
           options[:html][:novalidate] = !SimpleForm.browser_validations
         end
+
         if options[:html].key?(:class)
           options[:html][:class] = [SimpleForm.form_class, options[:html][:class]].compact
         else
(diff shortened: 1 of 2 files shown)


feliperenan:
Hi @KevinBongart and thanks for your contribution. :raised_hands: 

I don't like the idea to include a new option and it can lead to some confusion. Besides, you could achieve the same result if you put the desired class as default with `default_form_class` option + custom wrapper.

Thanks for the PR anyway. :+1:

bbugh:
@feliperenan this cannot be achieved in cases where forms need to have some kind of nested value in certain circumstances. The request for this functionality goes back years in many different stack overflow questions, many of which recommend using `html: { class: 'xxx' }`. It's very difficult to structure HTML effectively if we can't specify our own classes in different circumstances. Please reconsider this!

PR #1694, opened 2020-02-29
default_input_type cast to rich_text_area

mespina:


lib/simple_form/form_builder.rb
```diff
@@ -546,6 +546,7 @@ def default_input_type(attribute_name, column, options)
       return options[:as].to_sym if options[:as]
       custom_type = find_custom_type(attribute_name.to_s) and return custom_type
       return :select             if options[:collection]
+      return :rich_text_area     if defined?(ActionText::RichText) && @object.send(attribute_name).is_a?(ActionText::RichText)
```
rafaelfranca: This is getting the data from the database. We should not do this. If we should use reflection to know if an attribute is a rich tech but we should not serialize the data from the database in order to do that.

nashby:
Closing this for now since it's not an optimal way of doing it.

PR #1703, opened 2020-05-30
fix build_hidden_field_for_checkbox not returning ActiveSupport::SafeBuffer instance.

tfmcloughlin:
Returning a normal string here would result in raw html displaying on the final page, instead of the actual form. This fix ensure a ActiveSupport::SafeBuffer


nashby:
@tfmcloughlin thanks! I added a test case here https://github.com/heartcombo/simple_form/commit/21148b319bc9740af7580ecc44826df7bf5c5fcc

PR #1708, opened 2020-08-03
Explanation on how to customize the hint class.

silviorelli:
Added explanation on how to customize the hint class.


feliperenan:
Thanks 👍

PR #1722, opened 2021-01-05
I18nCache: Fix overtaken class var RuntimeError in Ruby 3.0

mcls:
This fixes errors like the one below in Ruby 3.0 by replacing the class variables with a singleton for caching the translations.

        RuntimeError: class variable @@translate_required_html of SimpleForm::Inputs::StringInput is overtaken by SimpleForm::Inputs::Base
          .../simple_form/lib/simple_form/i18n_cache.rb:13:in `class_variable_get'
          .../simple_form/lib/simple_form/i18n_cache.rb:13:in `get_i18n_cache'
          .../simple_form/lib/simple_form/i18n_cache.rb:8:in `i18n_cache'
          .../simple_form/lib/simple_form/components/labels.rb:9:in `translate_required_html'
          .../simple_form/lib/simple_form/components/labels.rb:71:in `required_label_text'
          .../simple_form/lib/simple_form/components/labels.rb:43:in `label_text'
          .../simple_form/lib/simple_form/components/labels.rb:35:in `label'

Related issue: https://github.com/heartcombo/simple_form/issues/1720

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rvm:
   - 2.5
   - 2.6
   - 2.7
+  - 3.0
 gemfile:
   - gemfiles/Gemfile.rails-5-0-stable
   - gemfiles/Gemfile.rails-5-1-stable
@@ -19,6 +20,12 @@ matrix:
     - rvm: 2.7
       gemfile: gemfiles/Gemfile.rails-5-1-stable
     - rvm: 2.7
+      gemfile: gemfiles/Gemfile.rails-5-2-stable
+    - rvm: 3.0
+      gemfile: gemfiles/Gemfile.rails-5-0-stable
+    - rvm: 3.0
+      gemfile: gemfiles/Gemfile.rails-5-1-stable
+    - rvm: 3.0
       gemfile: gemfiles/Gemfile.rails-5-2-stable
 before_install:
   - gem update --system
--- a/lib/simple_form/i18n_cache.rb
+++ b/lib/simple_form/i18n_cache.rb
@@ -5,19 +5,29 @@ module SimpleForm
   # caching facility to speed up form construction.
   module I18nCache
     def i18n_cache(key)
-      get_i18n_cache(key)[I18n.locale] ||= yield.freeze
-    end
-
-    def get_i18n_cache(key)
-      if class_variable_defined?(:"@@#{key}")
-        class_variable_get(:"@@#{key}")
-      else
-        reset_i18n_cache(key)
-      end
+      store = I18nCacheStore.instance
+      store[key] ||= {}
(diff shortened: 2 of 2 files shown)


pedrofurtado:
Pinging maintainers for notification (some of them may be on vacation) 🤝 : @rafaelfranca @tegon @feliperenan @carlosantoniodasilva

carlosantoniodasilva:
Just letting you know that I didn't forget about this, it's on my radar and I plan to work on it in the upcoming days alongside other Ruby 3 / Rails 6.1 upgrades. Thanks for the work here @mcls.

LeMinh-ThienToan:
I upgraded to Ruby 3.0.0, Rails 6.1.1 and got same error.

Thanks for the work.

carlosantoniodasilva:
@mcls I've been thinking some more about it, and considering dropping that cache entirely instead. It's extra complexity we've had to maintain over the years (and it's just growing a bit now), and I don't believe that cache is such a hot place with SimpleForm to be worth it the complexity.

That's where I'm leaning towards, I'll give it a couple more days. Let me know if you (or anyone else here) have any thoughts in the meantime. Thanks again!

mcls:
Hey, thanks for checking. Sounds good to me. 

I did a bit of digging and it looks like I18n also has its own [cache](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n/backend/cache.rb) which people can use if they find it necessary. I also think the default [YAML backend](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n/backend/simple.rb) is probably performant enough for most cases, given that it stores the translations in memory as hashes.

carlosantoniodasilva:
Sounds good, thanks for digging some more there @mcls. I'll get back to it in the next few days and prepare a new release.

carlosantoniodasilva:
Closed by https://github.com/heartcombo/simple_form/commit/05950880ba8750472a7fd62aba9ee79f49f530f6, forgot to reference this one too. Thanks!

matthieua:
@carlosantoniodasilva Awesome, thanks a lot for taking care of it! Do you think you can make a new release?

carlosantoniodasilva:
@matthieua thank you. Yes, definitely, this is on my radar for this week (as well as responders/devise), but I need to make sure https://github.com/heartcombo/simple_form/issues/1724 isn't a problem others will hit with SimpleForm when upgrading. (it doesn't appear to be SF specific, but more to investigate there.)

matthieua:
@carlosantoniodasilva no problem, that's what I thought, thanks for the update and the awesome work with SF!

pedrofurtado:
Thanks, @carlosantoniodasilva ! Happy to see a release with ruby 3 support asap 🎉  Thanks again! 🤝

PR #1725, opened 2021-01-28
Avoid a lot of NameError exceptions

kriom:
### Purpose

Greatly improves rendering time by avoiding the generation of `NameError` exception

### Context

Rendering a view with 966 inputs (ok I admit it's a mess) allowed me to see that this portion of code is not optimized at all.

I used https://github.com/tmm1/stackprof on my view to understand via the flamegraph that ~ 70% of the time is dedicated to handle `NameError` exceptions.

### Changes

Instead of trying to obtain a constant directly, it is verified that it exists first.

#### development environment :
Flamegraph  :arrow_left: before and after :arrow_right: fix :
Flamegraph before fix Flamegraph after fix
As you can see before the fix, a lot of time was spent in the `SimpleForm::FormBuilder#attempt_maping`.

In the development environment the view rendering speed up is ~=> X 5.0

#### production environment : _( edit: not reproduced afterwards )_

The same view goes from :
`Completed 200 OK in 2658ms (Views: 2637.2ms | ActiveRecord: 12.0ms | Allocations: 1240207)`
to
`Completed 200 OK in 1503ms (Views: 1482.9ms | ActiveRecord: 13.3ms | Allocations: 1244907)`

In the production environment the view rendering speed up is ~=> (2637 / 1482) = X 1.78 

And many thanks for your great work 😙

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -674,7 +674,7 @@ module SimpleForm
       return if SimpleForm.inputs_discovery == false && at == Object
 
       begin
-        at.const_get(mapping) if at.constants.include?(mapping.to_sym)
+        at.const_get(mapping) if at.const_defined?(mapping)
       rescue NameError => e
         raise if e.message !~ /#{mapping}$/
       end
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -674,7 +674,7 @@ module SimpleForm
       return if SimpleForm.inputs_discovery == false && at == Object
 
       begin
-        at.const_get(mapping)
+        at.const_get(mapping) if at.constants.include?(mapping.to_sym)
       rescue NameError => e
         raise if e.message !~ /#{mapping}$/
       end


lib/simple_form/form_builder.rb
```diff
@@ -674,7 +674,7 @@ def attempt_mapping(mapping, at)
       return if SimpleForm.inputs_discovery == false && at == Object
 
       begin
-        at.const_get(mapping)
+        at.const_get(mapping) if at.constants.include?(mapping.to_sym)
```
carlosantoniodasilva, line 677: Are you able to test this with `const_defined?`?

```suggestion
        at.const_get(mapping) if at.const_defined?(mapping)
```

I believe this might achieve similar results without the name error handling, but also without having to do sym conversion & array search.

carlosantoniodasilva (commented):
Thanks for your contribution, I had a suggestion below to test out if you can.

Are you testing this in production mode? And do you have `cache_discovery` enabled? Otherwise I think it will try to do the mapping lookups on every request (it should be on by default I think in production)

lib/simple_form/form_builder.rb
kriom: Thanks you for your response. Indeed it works perfectly with with `const_defined?`. It seems that we even gain a few milliseconds
development environment without fix **5105.5ms**
```
Completed 200 OK in 5133ms (Views: 5105.5ms | ActiveRecord: 17.7ms | Allocations: 2132426)
```
development environment first fix *895.0ms*
```
Completed 200 OK in 925ms (Views: 895.0ms | ActiveRecord: 17.1ms | Allocations: 746283)
```
development environment your fix *826.7ms*
```
Completed 200 OK in 853ms (Views: 826.7ms | ActiveRecord: 17.8ms | Allocations: 745690)
```

kriom:
I tested it in development and production modes but generated flamgraphs come from development mode
Rails log give me these speed up
- X 5.0 for development 
- X 1.78 for production

I use the default configuration for `cache_discovery`.
`config/initializers/simple_form.rb`
```
...
  # Cache SimpleForm inputs discovery
  # config.cache_discovery = !Rails.env.development?
...
```
In development `cache_discovery` is off
```
[3] pry(main)> Rails.env
=> "development"
[4] pry(main)> SimpleForm.cache_discovery
=> false
```
In production `cache_discovery` is on
```
irb(main):001:0> Rails.env
=> "production"
irb(main):002:0> SimpleForm.cache_discovery 
=> true
```

I will generate 2 flamegraphs without and with the fix in production mode

kriom:
Hello,

In production mode with and without the patch of this PR:
- the rendering durations are identical,
- the generated flamgraphs are identical.

I can no longer reproduce my performance improvement in production with only this patch. Which seems logical compared to what you say. It must have been due to another correction of my view (Oops).

So this improvement is only valid for the development mode.

carlosantoniodasilva:
Thanks @kriom, I believe that in production that discovery cache option takes place and helps prevent the many NameError rescues you were seeing in development.

I hope to circle back on this for review in the next few days.

cymen:
@carlosantoniodasilva I came across this issue while researching a similar `simple_form` rendering time issue and I found in our configuration, we had the default generated config that has the discovery cache option (that turns it off in non-development) commented out. Would a PR to change that so the option is not commented out in the configuration file be accepted?

I suspect many people are running `simple_form` in production with `cache_discovery` enabled and are not aware. The change I suggest above won't help them but at least new users won't experience that.

I was also surprised that I couldn't find any documentation for the discovery cache configuration setting. Maybe the impact isn't as dramatic in production but when I toggled it on in development, the admittedly complex form we are rendering took a bit less than 50% of the original time to render.

carlosantoniodasilva:
Thanks for your work on this one @kriom, and my apologies for not circling back on it back then. We got a similar change in https://github.com/heartcombo/simple_form/pull/1810, so I went ahead and merged that, credited both of you in the changelog.

@cymen I'm afraid that enabling that by default could have side-effects with code reloading in dev, i.e. if you make changes to input code it could remove the previous cached constant for example... this is off the top of my head, untested, but sounds like it would be possible. If you would like to give that a try and doesn't run into any issues, we could consider enabling that on by default in dev, but otherwise I think this should be fine as is. We could of course improve documentation on that config. (it should be on by default in production though.)

Thanks all!

carlosantoniodasilva:
I'm reverting the merged change, as it caused a regression. See #1824 for more info.

PR #1726, opened 2021-02-10
update bootstrap example app url

m5o:
Use direct heroku url since CNAME config doesn’t work anymore


carlosantoniodasilva:
Thanks!

PR #1731, opened 2021-03-23
bootstrap 5 update

yshmarov:
https://getbootstrap.com/docs/5.0/migration/#forms-2
Dropped .form-group for margin utilities (we’ve replaced our docs examples with .mb-3).


carlosantoniodasilva:
Please see https://github.com/heartcombo/simple_form/issues/1732#issuecomment-805751045 and https://github.com/heartcombo/simple_form-bootstrap/issues/323.

Thanks.

PR #1734, opened 2021-04-16
Add Truffleruby head to CI

gogainda:


carlosantoniodasilva:
I appreciate your contribution. Right now I'm not using other Ruby implementations other than MRI, and thus to avoid the maintenance overhead I'm not testing any heartcombo lib under these other Ruby implementations. Also, Rails itself does not run tests on those either. In other words, I'm assuming different MRI versions is going to be enough for now. This may change in the future and I can enable the tests for other rubies then, but for now I am not going to be able to add/maintain these. Thanks again.

PR #1738, opened 2021-05-28
Update install generator for Bootstrap 5

mhw:
This syncs up the generator template with the new one from `simple_form-bootstrap`.

There's a minor issue with the [styling of validation errors in input groups](https://github.com/heartcombo/simple_form-bootstrap/pull/334) that I'd like to look at, but I don't think we need to wait for that before merging this in.

/cc @m5o

### DIFF
--- a/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
+++ b/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
@@ -49,7 +49,7 @@ SimpleForm.setup do |config|
   # vertical forms
   #
   # vertical default_wrapper
-  config.wrappers :vertical_form, tag: 'div', class: 'form-group', error_class: 'form-group-invalid', valid_class: 'form-group-valid' do |b|
+  config.wrappers :vertical_form, class: 'mb-3' do |b|
     b.use :html5
     b.use :placeholder
     b.optional :maxlength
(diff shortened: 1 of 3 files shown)


m5o:
👀 https://github.com/heartcombo/simple_form/pull/1729 maybe could become interesting with the issue around `.has-validation` for input groups

oTiMo:
Thank you for this PR !

lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
```diff
@@ -49,98 +49,108 @@
...
+      bb.use :hint, wrap_with: { class: 'form-text' }
     end
   end
 
   # vertical input for radio buttons and check boxes
-  config.wrappers :vertical_collection, item_wrapper_class: 'form-check', item_label_class: 'form-check-label', tag: 'fieldset', class: 'form-group', error_class: 'form-group-invalid', valid_class: 'form-group-valid' do |b|
+  config.wrappers :vertical_collection, item_wrapper_class: 'form-check', item_label_class: 'form-check-label', tag: 'fieldset', class: 'mb-3' do |b|
     b.use :html5
     b.optional :readonly
     b.wrapper :legend_tag, tag: 'legend', class: 'col-form-label pt-0' do |ba|
```
davekaro, line 82: I found with bootstrap 5, the reboot css has `float: left` applied to `legend` elements. See https://github.com/twbs/bootstrap/pull/30345. It adds a `clear: left` to the next element, but in rails adds an auxiliary hidden field before every collection of radio buttons, which the `clear: left` gets applied, but since it's hidden has no effect. I fixed this by adding `float-none` class to this list, though I'm not sure that is the best solution.

mhw: Yes, I found the same issue when I was preparing this in the `simple_form-bootstrap` project. I addressed it with an [additional css rule](https://github.com/heartcombo/simple_form-bootstrap/blob/83904588bf3da26647092d1ce85742b170d27398/app/assets/stylesheets/simple_form-bootstrap/_form_legend_clear.scss) that targets the first following `div.form-check`. I'd wondered about submitting this upstream but haven't done so yet.

oTiMo:
When do you think this branch will be merged?

collimarco:
Nice to find this issue... I was really wondering what was wrong with my configuration, since the design was broken with Bootstrap 5 in a new app. 

It seems that copying the new initializer solves most issues, Thanks!! 

The only strange thing is that the `btn` is completely white (and it doesn't look like a button), maybe it should be set to `btn-primary` by default?

jamgregory:
I've just come across this issue when generating a new Rails application that uses Bootstrap 5 and realising that Simple Form doesn't have Bootstrap 5 support yet.

@carlosantoniodasilva - I don't suppose you know when this might be merged in and released? :crossed_fingers:

nashby:
@mhw thank you!

PR #1751, opened 2021-11-26
Multiple Wrapper Mappings

braindeaf:
Allow usage of wrapper_mapping sets whilst maintaining backwards
compatible config as per issue/suggestion
https://github.com/heartcombo/simple_form/issues/1632

### DIFF
--- a/lib/simple_form.rb
+++ b/lib/simple_form.rb
@@ -139,11 +139,21 @@ See http://blog.plataformatec.com.br/2019/09/incorrect-access-control-in-simple-
 
   # Custom wrappers for input types. This should be a hash containing an input
   # type as key and the wrapper that will be used for all inputs with specified type.
-  # e.g { string: :string_wrapper, boolean: :boolean_wrapper }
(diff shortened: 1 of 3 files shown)


braindeaf:
We've been using my fork in production for just over a year now :)

ldonnet:
It could be so helpful if this PR could be merged because it could be very useful in many use cases.  On my side, I want to migrate from bootstrap to tailwind with specific simple form wrappers and this PR permits to improve gem uses.
@braindeaf Thanks for your work and explanations!

ldonnet:
@braindeaf Have you continue improvements like integrated wrapper_mapping in form ?

```ruby
config.wrappers :my_special_form do |b|
  b.wrapper_mapping :my_special_mapping
  b.use :input
  # ...
end

config.wrapper_mappings[:my_special_mapping] = {
  boolean: :my_special_boolean,
  # ....
}
```

Then when using the form in the view like this it would know which form wrapper to use and also automatically which mapping to use for this wrapper:

```ruby
simple_form_for(animal, wrapper: :my_special_form)
```

PR #1763, opened 2022-02-23
Add Ruby 3.1 to the CI matrix

petergoldstein:
Adds Ruby 3.1 to the CI matrix.  In addition to the minimal change to the CI configuration, the following additional changes are made to support the addition:

1. The version of the Rails in `Gemfile` and `Gemfile.lock` is updated from `7.0.0.rc1` to `7.0.2.2`
2. Additional gem versions in `Gemfile.lock` are updated via `bundle update`
3. The `CHANGELOG.md` is updated to reflect the above
4. The `3.0` in the CI configuration is quoted, so that it properly resolves to a 3.0.x version.  An unquoted 3.0 will be truncated to 3, causing it to load the latest Ruby 3 - at this time 3.1.1.  To ensure a 3.0.x version is loaded, we quote the 3.0.


petergoldstein:
Note that the 3.0 is being truncated to 3, and the version of Ruby being loaded is 3.1.1:

Screen Shot 2022-02-23 at 7 43 21 AM

PR #1764, opened 2022-03-09
Update bootstrap initializer banner

### LINKED ISSUE #1756: Change the Bootstrap classes used: collision between the initializers
Run:

```
rails generate simple_form:install --bootstrap
```

Now you have two initializers.

Suppose that you want to set this:

```
config.button_class = 'btn btn-lg btn-primary px-5'
```

**Option 1 - Set it in Bootstrap initializer**

You should not do that, as documented in the file:

```
Please do not make direct changes to this file!
```

**Option 2 - Set it in default initializer**

In this case setting the value has not effect, because the other initializer is taking over.

**Basically, if you follow the official instructions, there is no way to customize the Bootstrap forms :(**

m5o:
* remove a misleading sentence
  * `Please do not make direct changes to this file!`
* 💭 prep & discussion:
  * https://github.com/heartcombo/simple_form-bootstrap/pull/359

fixes #1756

/cc @carlosantoniodasilva @collimarco

### DIFF
--- a/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
+++ b/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb
@@ -1,10 +1,8 @@
 # frozen_string_literal: true
 
-# Please do not make direct changes to this file!
-# This generator is maintained by the community around simple_form-bootstrap:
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
@m5o thank you

PR #1765, opened 2022-03-11
Only call where if needed

asgerb:
Calling `where` with a nil argument is not supported in [mongoid](https://github.com/mongodb/mongoid/blob/5c90efb00347e25d8b5ff0ce27d00ac8c906f0e2/lib/mongoid/criteria/queryable/selectable.rb#L795) and throws an error.

Calling `where` only if there are any conditions fixes this issue and seems to me a straightforward change.
The question is if there are cases where you want to call `where` with a nil argument?


nashby:
@asgerb thanks! Totally makes sense.

PR #1766, opened 2022-04-11
Update README.md to fix the end_year value

gurudath:
The end date value was not meaningful full to have Date.today.year - 12 it should be Date.today.year + 12 on the README.md.


carlosantoniodasilva:
Appreciate the contribution.

The sample is a `date_of_birth` field, you can't really have one in the future so adding to the current year doesn't really work for that example.

PR #1772, opened 2022-05-10
Honor `include_hidden` on inline boolean inputs

yboulkaid:
On `nested`-style boolean inputs, it's possible to opt-out of the hidden input
generation by passing the `include_hidden: false` option.

However, for `inline` inputs, this still generates the hidden field, causing
an inconsistency between the two boolean styles, and a whole lot of late night confusion.

A workaround has been proposed in [a comment from 2016](https://github.com/heartcombo/simple_form/issues/1377#issuecomment-227616280),
which consists in passing `include_hidden` in the `input_html`. I think this is 
not ideal, as it causes inconsitencies between the APIs of the boolean styles.

This PR makes the inline checkboxes honor the `include_hidden` styles, making their API consistent.

### DIFF
--- a/lib/simple_form/inputs/boolean_input.rb
+++ b/lib/simple_form/inputs/boolean_input.rb
@@ -12,7 +12,11 @@ module SimpleForm
                 inline_label
             }
         else
-          build_check_box(unchecked_value, merged_input_options)
+          if include_hidden?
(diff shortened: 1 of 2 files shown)


nashby:
@yboulkaid totally makes sense, thanks!

PR #1779, opened 2022-06-28
Add Dependabot for GitHub Actions

petergoldstein:
This PR adds Dependabot for GitHub Actions as described here - https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

As the checkout action is currently out of date, Dependabot will open a PR to update it if and when this PR is merged.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request but I don't think we need need to update those things often to require a dependabot config.

Issue #1780, opened 2022-06-29
docs query - is example with "unchecked_value: false" still appropriate?


paulcc:
I think an example in the README docs needs to be revised. 

I'm not sure what's best - depends what you're wanting to show with this example. Maybe the best option is to use strings and to pick ones which are clearly app-specific, eg use "alpha" and "beta" as the options, for cases when you want to diverge from usual Rails handling of check boxes.  

See https://github.com/heartcombo/simple_form/blame/main/README.md#L260

```
f.input :accepts, as: :boolean, checked_value: true, unchecked_value: false
```

This example was last changed 9 years ago, but the behaviour of unchecked_value has changed since, eg https://github.com/heartcombo/simple_form/commit/dfcd041b224fb3ae4d748bd0c51b7588016e19eb in 2017. 

Note that if you use the example as above, you currently will not get the hidden field created for the checkbox, which is probably not what the example is trying to show. (We updated an old app recently, and got caught by this!)

nashby:
@paulcc yep, you're right! Can you please send a PR that changes `true` and `false` to some strings?

nashby:
closed by #1784

PR #1784, opened 2022-07-12
Revise unchecked docs

paulcc:
Addresses issue #1780. 

Two commits:
 * simple string change
 * new comment clarifying the handling of hidden field on check boxes


nashby:
thanks @paulcc!

Issue #1638, opened 2019-01-30
rich_text_area input field type for ActionText shipping with Rails 6.x


msdundar:
## Environment

- Ruby **[version]**

ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-linux]

- Rails **[version]**

6.x

- Simple Form **[version]**

simple_form (4.1.0)

## Current behavior

Rails 6 ~is close to shipping~ has released. ActionText is one of the most exciting features of Rails 6, and it is implemented to Rails forms as follows:

```ruby
<%= form_with(model: message) do |form| %>
    <%= form.label :content %>
    <%= form.rich_text_area :content %>
<% end %>
```

However, `simple_form` doesn't seem ready to accept `rich_text_area` as an argument to `as:`.

## Expected behavior

`simple_form` should support `rich_text_area` as an input type and render the ActionText form.

feliperenan:
I'm not sure if we should include this feature at Simple Form now. @rafaelfranca any thoughts?

alec-c4:
Hi, lads! Any news about this issue?

aruljohn:
Is `simple_form` going to support `rich_text_area` for Rails 5.2?

bopm:
So Rails 6 is out in the wild, but no news here.

jaredmoody:
In the meantime, this is pretty trivial to implement yourself:

```
class RichTextAreaInput < SimpleForm::Inputs::Base
  def input(wrapper_options = nil)
    merged_input_options = merge_wrapper_options(input_html_options, wrapper_options)

    @builder.rich_text_area(attribute_name, merged_input_options)
  end
end
```

ThomasConnolly:
Trivial for you, Jared...
...but I don't know where I would put that code. In an initializer, such as simple_form.rb? I don't understand the reluctance to update the gem to be used with action_text, especially for blokes like me who only know only enough RoR to operate a single website as a volunteer.

jaredmoody:
@ThomasConnolly where to put the code is more a Rails topic than SimpleForm, but I have my custom inputs in `app/inputs`; since `app` is in the Rails default autoload path, there's nothing more to do.

You could also put custom inputs under `lib/inputs` perhaps, in which case you would need to add a `require` statement to the file(s) - or add `lib/inputs` to `config.autoload_paths` in `config/application.rb` so that rails will autoload everything in that folder.

[...]

PR #1793, opened 2022-10-14
Add missing documentation for ActionText

r-sierra:
Hi there!, first of all I wanna thank everyone for this great gem :rocket: . 

I was trying to use this gem with an input defined with `has_rich_text` but could not find any reference on how to do it. Found this issue #1638, and near the end of the discussion found some comments, with suggestions on how to use it. So I decided to add this missing documentation for future references.


MatheusRich:
@nashby is there anything actionable here?

calebwoods:
Just discovered this exists as well. Would be great to merge this in so the documentation is up to date on what is supported out of the box

nashby:
@r-sierra thanks! I've pushed https://github.com/heartcombo/simple_form/commit/62149be7f093ed67cdb23a5a7a4eaaee6946b9b8 instead of your change since we don't map `rich_text_area` to any DB type currently. Potentially we can check if model responds to `rich_text` methods but we're not doing it now.

PR #1794, opened 2022-11-18
Use label or legend tag wrapper to be consistent

martinodonnell:
Address https://github.com/heartcombo/simple_form/issues/1788

Goal is to make the bootstrap templates consistent on how labels are created


martinodonnell:
Looking at the file, I should proabbly move this to https://github.com/heartcombo/simple_form-bootstrap. This right?

https://github.com/martinodonnell/simple_form/blob/fa714cbde02dab523dd75558dc693400f18a08da/lib/generators/simple_form/templates/config/initializers/simple_form_bootstrap.rb#L3-L5

Issue #1723, opened 2021-01-15
Some integration support for Tailwindcss?


DiegoOrejuela:
Hi guys

I would like to ask if you are considering adding support to Tailwindcss

Thanks

carlosantoniodasilva:
I don't have experience with Tailwind to build anything out of the box, it's not something that's on my radar for the near term at least, but perhaps in the future.

I do think that the wrapper support in SF should be mostly enough to build a set of default configs for your own needs, and I'd be happy to support adding anything that might be missing, but I'm not sure about maintaining such configs with SF itself for now. If you (or someone else coming here) have a set of configs, maybe a starting point would be the wiki.

FWIW, Bootstrap support was added really long ago and is now mainly maintained via https://github.com/rafaelfranca/simple_form-bootstrap, and there hasn't been any changes for the past couple of years.

abevoelker:
As mentioned in #1727, I made an attempt at adding a Tailwind component using the Wrappers API:

```ruby
config.wrappers :tailwind_text_input, tag: 'div', class: '', error_class: '', valid_class: '' do |b|
  b.use :html5
  b.use :placeholder
  b.optional :maxlength
  b.optional :minlength
  b.optional :pattern
  b.optional :min_max
  b.optional :readonly

  b.use :label, class: "block text-sm font-medium text-gray-700"
  b.wrapper tag: 'div', class: 'mt-1', error_class: 'relative rounded-md shadow-sm' do |c|
    c.use :input,
      class: 'appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-indigo-500 focus:border-indigo-500 sm:text-sm',
      error_class: 'block w-full pr-10 border-red-300 text-red-900 placeholder-red-300 focus:outline-none focus:ring-red-500 focus:border-red-500 sm:text-sm rounded-md'
    c.use :full_error, wrap_with: { tag: 'p', class: 'mt-2 text-sm text-red-600' }
  end
end
```

Which I use like this:

```ruby
<%= f.input :email, wrapper: :tailwind_text_input, autocomplete: "email", placeholder: "asmith@example.com", label: "Email address" %>
```

That wrapper is a first attempt to mimic this HTML:

```ruby
<%= simple_form_for(@user, url: profiles_path, html: { class: "space-y-6" }) do |f| %>
  <div>
    <%= f.label :email, class: "block text-sm font-medium text-gray-700" do %>
[...]

PR #1796, opened 2022-11-23
Add default_builder configuration option

gregorw:
For some CSS frameworks the current Simple Form configuration options are not enough. In particular, the error notification may require [additional nesting](https://github.com/abevoelker/simple_form_tailwind_css/blob/master/lib/simple_form/tailwind/error_notification.rb). But also data attributes on some wrappers may be [required](https://react.carbondesignsystem.com/?path=/story/components-textinput--playground&args=invalid:true). By introducing the `default_builder` configuration option any framework should be able to be configured.

In the case of [simple_form_tailwind_css](https://github.com/abevoelker/simple_form_tailwind_css#usage) this would allow to use

```erb
<%= simple_form_for(@foo) do |f| %>
```

instead of

```erb
<%= simple_form_for(@foo, builder: SimpleForm::Tailwind::FormBuilder) do |f| %>
```

See also #1723.

### DIFF
--- a/lib/generators/simple_form/templates/config/initializers/simple_form.rb
+++ b/lib/generators/simple_form/templates/config/initializers/simple_form.rb
@@ -173,4 +173,7 @@ SimpleForm.setup do |config|
   # Defines validation classes to the input_field. By default it's nil.
   # config.input_field_valid_class = 'is-valid'
   # config.input_field_error_class = 'is-invalid'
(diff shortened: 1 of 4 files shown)


pas256:
This looks great to me, I don't see why it can't be merged in. Perhaps it needs some documentation on the README?

PR #1810, opened 2023-02-20
Speed up nested #simple_fields_for usage by a considerable amount

meanphil:
The `FormBuilder#find_mapping` method is relying on `Object#const_get` raising an exception if the constant doesn't exist, rather than directly checking if it exists with `Object#const_defined?`.

Exceptions are slow due to generating backtraces - and then when other debugging gems (such as did_you mean, which are included with Ruby and Rails these days) get involved it resulted in excessive Object allocation.

Normally the `discovery_cache` would mitigate this, so there wouldn't be many calls to `#mapping_override`, however when you use `#simple_fields_for` with a collection/has_many association the cache is empty for each item in the collection.

Before this change this code was taking approximately 2 seconds and 2.9million allocations for 235 records:

```
<%= f.simple_fields_for :provider_access_masks, @access_masks do |pam| %>
  <%= pam.input :id, as: 'hidden' %>
  <%= pam.input :_destroy, as: 'boolean', label: "Delete", class: 'destroy' %>
<% end %>
```

After this change it was down to approximately 100ms and 200,000 allocations. (All on my M1 MacBook Air)

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -672,12 +672,7 @@ module SimpleForm
 
     def attempt_mapping(mapping, at)
       return if SimpleForm.inputs_discovery == false && at == Object
-
-      begin
-        at.const_get(mapping)
-      rescue NameError => e
-        raise if e.message !~ /#{mapping}$/
-      end
+      return at.const_get(mapping) if at.const_defined?(mapping)
     end
 
     def attempt_mapping_with_custom_namespace(input_name)


carlosantoniodasilva (approved):
I had a recollection of seeing something like this before, and sure enough there was a similar PR a while ago: https://github.com/heartcombo/simple_form/pull/1725, I just haven't been able to circle back on that one at the time.

This all sounds like a safe change, and hopefully an improvement to the slowness caused by raising/rescuing exceptions in general. Appreciate it!

carlosantoniodasilva:
I'm reverting this change, as it caused a regression. See #1824 for more info.

PR #1812, opened 2023-03-01
Pass base object to `human_attribute_name` for labels

motine:
When discovering translations, Rails [passes the `base` object to `human_attribute_name`](https://github.com/rails/rails/blob/af09d6b90b23c2d7389f879f37e9cbbeea078e5a/activemodel/lib/active_model/error.rb#L55). This is helpful because `human_attribute_name` is a class method and does not have access to instance methods. If someone wants to customize the behavior of said method based on properties of the instance, this is not possible without this argument. This merge request makes a tiny change to pass this base object.


nashby:
@motine thanks, that's useful indeed!

PR #1823, opened 2023-10-31
Remove aria-required for required fields in HTML5

### LINKED ISSUE #1822: Redundant aria-required on required fields
## Environment

- Ruby **3.2.2**
- Rails **7.0.0**
- Simple Form **5.3.0**

## Current behavior

This form:
```
<%= simple_form_for(@user, url: simple_form_test_url) do |f| %>
    <%= f.input :uuid, required: true %>
<% end %>
```
produces this output:
```
<input class="required" required="required" aria-required="true" type="text" value="333...678" name="user[uuid]" id="user_uuid">
```
Having both `required="required"` and `aria-required="true"` is redundant and not recommended [per the Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required#examples).

## Expected behavior

Ideally, the generated field would look like this:
```
<input class="required" required="required" type="text" value="333...678" name="user[uuid]" id="user_uuid">
```
We see that this behavior was added intentionally in #781 (2013), but is no longer recommended.

aduth:
Closes #1822

This pull request proposes to remove the inclusion of the `aria-required` attribute on required fields, since the fields are already assigned the `required` attribute which would be used by assistive technology to convey the field requirement. See #1822 for additional context.


nashby:
@aduth thank you!

Issue #1739, opened 2021-05-28
Generate HTML5 date/time inputs by default


marcelolx:
It's a long time ago since HTML5 date/time support was included in simple_form https://github.com/heartcombo/simple_form/pull/935 and since then as the docs describe by default they are disabled https://github.com/heartcombo/simple_form/pull/1150 because at the time the browsers weren't ready for them, but I think now 8 years after, the browsers are ready as mdn browser compatibility suggests https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#browser_compatibility, what do you think about simple_form turn on the generation of these html5 inputs by default?

This probably would be a breaking change and would require a major release, I suppose

attenzione:
This can be implemented without breaking change, but just add another config option which is `false` by default:
```ruby
# initializer
SimpleForm.setup do |config|
  config.default_html5_date_time = false
end

# input helper (simple_form/inputs/date_time_input.rb#L34)
def use_html5_inputs?
  input_options.fetch(:html5, SimpleForm.default_html5_date_time || false)
end
```

nashby:
It looks like a bug to be honest. https://github.com/heartcombo/simple_form/blob/main/lib/simple_form/inputs/date_time_input.rb#L35 it doesn't use `html5` config from the wrapper but uses `html5` option from the input options. @carlosantoniodasilva do you thing we should make it work the same as e.g [NumericInput](https://github.com/heartcombo/simple_form/blob/main/lib/simple_form/inputs/numeric_input.rb#L9) works?

rafaelfranca:
@nashby yes. I think we should match the behavior.

jtarchie:
I just experienced this. For anyone reading, you have explicitly set `html5: true, as: :date` on the field to have it render correctly.

jamezilla:
I opened a PR for this. It's not a breaking change, but it's definitely a change to the default behavior since the generated [default wrapper](https://github.com/heartcombo/simple_form/blob/main/lib/generators/simple_form/templates/config/initializers/simple_form.rb#L26) is using the `html5` component.

PR #1830, opened 2024-02-14
Fix DatetimeInput so it respects html5 wrapper configuration

jamezilla:
Bug fix for #1739

I reorganized the tests a little bit. Tests which should generate html5 inputs were consolidated under `DateTimeInputWithHtml5Test` and tests which shouldn't generate html5 inputs were consolidated under `DateTimeInputWithoutHtml5Test`.

I also removed some unnecessary calls to `swap_wrapper` since the default wrapper supports optional use of the html5 component.

### DIFF
--- a/lib/simple_form/inputs/date_time_input.rb
+++ b/lib/simple_form/inputs/date_time_input.rb
@@ -5,7 +5,7 @@ module SimpleForm
       def input(wrapper_options = nil)
         merged_input_options = merge_wrapper_options(input_html_options, wrapper_options)
 
-        if use_html5_inputs?
+        if html5?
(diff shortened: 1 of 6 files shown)


csculley:
Is there anything beyond updating this PR to the latest commit that is preventing this PR from getting merged? I'm interested in seeing this behavior :smile:

In https://github.com/heartcombo/simple_form/issues/1739#issuecomment-1089484688, @rafaelfranca mentioned that he'd like to see the existing behavior matched, which might require additional changes to this ticket and the default configuration to disable these html5 fields by default.

This PR could probably also be updated to say `Fixes #1739` or have the issue added manually by a moderator since it doesn't seem to be linked to #1739

PR #1833, opened 2024-03-05
Pin country_select version in README

DanielPechersky:
The latest release of Simple Form breaks with `country_select` 9.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ If you want to use the country select, you will need the
 [country_select gem](https://rubygems.org/gems/country_select), add it to your Gemfile:
 
 ```ruby
-gem 'country_select'
+gem 'country_select', '~> 8.0.3'
 ```
 
 If you don't want to use the gem you can easily override this behaviour by mapping the


DanielPechersky:
I'll get you more information on the error, but you should pin the version of `country_select` anyways. Otherwise the user's code will break in this circumstance:
1. A user copy pastes `gem 'country_select'` from your README
2. `country_select` publishes a new major version
3. The user updates all their gems, and `country_select` is no longer compatible with Simple Form

carlosantoniodasilva:
I appreciate the feedback. How users decide to define the gem on their gemfiles is probably not a big concern we should have though, some prefer to pin, others prefer to leave it open, in any case bundle updating multiple gems is generally not advised.

There's also the concern of it becoming an extra small burden to be updating the readme every time with new `country_select` versions ourselves. Plus, we wouldn't get these helpful reports when it breaks. :)

DanielPechersky:
This seems to be the relevant information:

```
#<ArgumentError: wrong number of arguments (given 4, expected 1..3)>

/Users/dannyp/.rbenv/versions/3.2.2/gemsets/foo/gems/country_select-9.0.0/lib/country_select/country_select_helper.rb:6:in `country_select'
/Users/dannyp/.rbenv/versions/3.2.2/gemsets/foo/gems/formtastic-5.0.0/lib/formtastic/inputs/country_input.rb:77:in `block in to_html'
```

carlosantoniodasilva:
@DanielPechersky it looks like you're using [formtastic](https://github.com/formtastic/formtastic) instead, which has a [PR to add support to country_select v9.0.0](https://github.com/formtastic/formtastic/pull/1382).

Let me know if you run into anything with SimpleForm.

DanielPechersky:
Ah, looks like this is the fabled "user error". Apologies

carlosantoniodasilva:
No worries, glad I could help. 👍

PR #1836, opened 2024-03-07
Fix encrypted attributes improperly casted

Michaelvilleneuve:
This PR fixes a bug introduced by ActiveRecord::Encrypted attributes. 
Instead of returning a `ActiveModel::Type::String` like they used to, encrypted attributes return an instance of `ActiveRecord::Encryption::EncryptedAttributeType`, on which you need to call `#cast_type` in order to obtain the original type. 

To fix this, I check whether or not the attribute `#responds_to?(:cast_type)` which would indicate that we need to take into account the underlying type instead of just `#type_for_attribute`.

Checking this instead of checking if the `attribute_name` `#responds_to?(:encrypted_attributes)` allows to take into account other forms of type overriding that ActiveRecord may introduce in the future or potentially already exist. 

See the type difference below between a regular string attribute and an encrypted attribute
image

Fixes https://github.com/heartcombo/simple_form/issues/1829

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -599,7 +599,11 @@ module SimpleForm
 
     def find_attribute_column(attribute_name)
       if @object.respond_to?(:type_for_attribute) && @object.has_attribute?(attribute_name)
-        @object.type_for_attribute(attribute_name.to_s)
+        detected_type = @object.type_for_attribute(attribute_name.to_s)
+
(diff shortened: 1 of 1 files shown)


nashby:
@Michaelvilleneuve hey! Thanks for the PR! It looks like Rails bug to me so I've sent a PR fixing it https://github.com/rails/rails/pull/52247 but I think it's ok to merge this for now.

carlosantoniodasilva:
@diesl we'll look into a release soon.

PR #1837, opened 2024-03-19
Allow to give an object's method name as symbol define collection

kriom:
Hi,

I would like to be able to give a method as symbol for the collection
with, for example
```ruby
class User
  #...
  def locales
    [:"en-US", :"fr-FR", :"es-ES", :"ja-JP", :"zh-CN"]
  end
  #...
end
```
I would like to do that in my view
```ruby
form.input(:locale, collection: :locales, include_blank: false)
```
rather than :
```ruby
form.input(:locale, collection: form.object.locales, include_blank: false)
```
Is it feasible? ?

### DIFF
--- a/lib/simple_form/inputs/collection_input.rb
+++ b/lib/simple_form/inputs/collection_input.rb
@@ -33,7 +33,12 @@ module SimpleForm
 
       def collection
         @collection ||= begin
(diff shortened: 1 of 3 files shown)


kriom:
OK no problem, thanks for the answer

PR #1845, opened 2024-06-30
Update Ruby/Rails support

nashby:
* Add support for Ruby 3.4 and Rails 7.2/8.0/8.1.
* Drop support for Rails < 7 and Ruby < 2.7.


carlosantoniodasilva:
Thanks @nashby! I went ahead and updated with some more versions dropping and a few other cleanup bits.

PR #1847, opened 2024-07-17
Fix small mistake in README.md

francois-ferrandis:
As far as I know: 
- to prepend is to add at the beginning
- to append is to add at the end

Simple Form adds an asterisk at the **end** of the label for required fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ any html attribute to that wrapper as well using the `:wrapper_html` option, lik
 <% end %>
 ```
 
-Required fields are marked with an * prepended to their labels.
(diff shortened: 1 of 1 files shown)


nashby:
Prepend is what SimpleForm does by default, see https://github.com/heartcombo/simple_form/blob/c2c7faf3e532072e1ef511a97d677ae0e7c38957/lib/simple_form.rb#L104

but you can configure it any way you like (that's probably what you have in your app)

francois-ferrandis:
Woops, sorry, I should have been more thorough. :grimacing: 

Thank you for your time! :pray:

PR #1861, opened 2026-02-07
Fix required on belongs_to for rails 7.1+

### LINKED ISSUE #1848: `belongs_to` relationships do not show up as required on rails 7.1+
## Environment

- Ruby **3.3.0**
- Rails **7.2.0**
- Simple Form **5.3.1**

## Current behavior

By default, on rails 7.1+, `belongs_to` relationships do not show up as required.

<details>
<summary>Sample App</summary>

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'rails'
  gem 'simple_form'
  gem 'sqlite3'
end

require 'minitest/autorun'
require 'active_record'
require 'action_view'

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new($stdout)

ActiveRecord::Schema.define do
  create_table :users, force: true
  create_table :payments, force: true do |t|
    t.references :user
  end
end

class User < ActiveRecord::Base
end

class Payment < ActiveRecord::Base
  belongs_to :user
end

class RequiredTest < ActionView::TestCase
  def test_required
    render inline: <<~ERB
      <%= simple_form_for Payment.new, url: '' do |f| %>
        <div>
          <%= f.association :user %>
        </div>
      <% end %>
    ERB

    element = rendered.html.at('div')

    assert_match /\*/, element.text
  end
end
```

This works if we add a redundant `validates :user, presence: true`.

</details>

## Expected behavior

`belongs_to` relationships should show up as required

## Details

This is due to an interaction between how we calculate required, specifically at https://github.com/heartcombo/simple_form/blob/c2c7faf3e532072e1ef511a97d677ae0e7c38957/lib/simple_form/helpers/validators.rb#L24, and a new change in rails 7.1 (https://github.com/rails/rails/blob/9f9deaf6c14e3fdb3134261a9d52c57208c750ee/activerecord/lib/active_record/associations/builder/belongs_to.rb#L127-L140 / https://github.com/rails/rails/pull/46522) to conditionally define the validation.

In the short term, we can fall back to the old behavior by setting [config.active_record.belongs_to_required_validates_foreign_key](https://guides.rubyonrails.org/configuring.html#config-active-record-belongs-to-required-validates-foreign-key) to true (thus unconditionally defining the validation), but my understanding is that this is a short-lived config to support backwards compatibility, and besides, the new behavior is a performance improvement.

Assuming y'all agree this is something that should be fixed (rather than, say, defining the redundant validation, passing `required: true` to the field in the form, or something else), I'd be happy to work on a patch if I could get some guidance. Something like

```diff
-  validator.options.include?(:if) || validator.options.include?(:unless) 
+ if validator.options.include?(:if)
+   file_path, line_number = validator.options[:if].source_location
+   !file_path.match?(%r{lib/active_record/associations/builder/belongs_to.rb}) && line_number != 130
+ else
+   validator.options.include?(:unless)
+ end
```

would obviously work but be incredibly hacky. How would we want to approach a more robust way to allow this particular conditional validation where the item should always be present?

fa11enangel:
close #1848

I've added the missing test cases based on ValidatingUser model and fixed them in validators.rb. I'm not sure I got all the cases, but this one is working for my projects.

### DIFF
--- a/lib/simple_form/helpers/validators.rb
+++ b/lib/simple_form/helpers/validators.rb
@@ -21,7 +21,46 @@ module SimpleForm
       end
 
       def conditional_validators?(validator)
-        validator.options.include?(:if) || validator.options.include?(:unless)
+        if_condition = validator.options[:if]
+        unless_condition = validator.options[:unless]
+
+        return false if if_condition.nil? && unless_condition.nil?
+
+        conditional = false
+
+        if if_condition
+          conditions = Array(if_condition)
+          conditional = conditions.any? { |c| !resolve_if_condition(c) }
+        end
+
+        if !conditional && unless_condition
+          conditions = Array(unless_condition)
+          conditional = conditions.any? { |c| resolve_if_condition(c) }
+        end
+
+        conditional
+      # If evaluating the condition fails (e.g., the condition references a method
+      # the object doesn't have, or a callable with unexpected arity), fall back to
+      # treating the validator as conditional and skipping it.
+      rescue NoMethodError, ArgumentError
+        true
+      end
+
+      def resolve_if_condition(condition)
+        case condition
+        when Symbol
+          object.send(condition)
+        else
+          if condition.respond_to?(:call)
+            if condition.is_a?(Proc) && condition.arity == 0
+              condition.call
+            else
+              condition.call(object)
+            end
+          else
+            condition
+          end
+        end
       end
 
       def action_validator_match?(validator)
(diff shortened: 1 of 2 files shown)


fa11enangel:
Thank you for your explanation @carlosantoniodasilva . These are all very valid concerns.

@ramaboo you can use my changes and use a fork of simple_form like I do for my projects.

PR #1866, opened 2026-04-19
Fix collection option being ignored for priority inputs (time_zone and country)

### LINKED ISSUE #1851: [Bug] Collection option for time zone input doesn't limit the selection
## Environment

- Ruby **3.2.2**
- Rails **7.0**
- Simple Form **5.3.1**

## Code sample

``` ruby
f.input :time_zone, as: :time_zone, collection: ["Eastern Time (US & Canada)", "Pacific Time (US & Canada)", "London", "Berlin"].map { ActiveSupport::TimeZone[_1] }, include_blank: false
```

## Current behavior

Renders all time zones as options

## Expected behavior

Renders only specified time zones selection.

55728:
Fixes #1851

### Problem

When `collection:` was passed to a `time_zone` or `country` input, it was silently ignored — all options were still rendered.

This also affects country inputs, as [reported by @Tressa-Sanders](https://github.com/heartcombo/simple_form/issues/1851#issuecomment-2345871424).

### Root cause

`PriorityInput#input` always delegates directly to `time_zone_select` / `country_select`, bypassing the parent `CollectionSelectInput#input` which is responsible for rendering the `collection:` option.

### Solution

Delegate to `super` when `collection:` is present, so the built-in collection rendering in `CollectionSelectInput` takes over.

### Known limitation

When `collection:` is given, `priority:` (and the global `SimpleForm.time_zone_priority` / `SimpleForm.country_priority` configuration) will not produce priority separators, because `CollectionSelectInput` uses `collection_select` rather than the native `time_zone_select` / `country_select` helpers. This matches the natural division of responsibilities: `collection:` opts into generic collection rendering.

Supporting both simultaneously would require rethinking the `PriorityInput` architecture and could be addressed in a follow-up.

### Tests added

- `collection:` with `[label, value]` pairs (time_zone + country)
- `collection:` with `ActiveSupport::TimeZone` objects (reproduces the exact code from #1851)

All 809 tests pass.

### DIFF
--- a/lib/simple_form/inputs/priority_input.rb
+++ b/lib/simple_form/inputs/priority_input.rb
@@ -3,6 +3,8 @@ module SimpleForm
   module Inputs
     class PriorityInput < CollectionSelectInput
       def input(wrapper_options = nil)
+        return super if options[:collection]
+
         merged_input_options = merge_wrapper_options(input_html_options, wrapper_options)
 
         send(:"#{input_type}_input", merged_input_options)
--- a/test/inputs/country_input_test.rb
+++ b/test/inputs/country_input_test.rb
@@ -30,6 +30,14 @@ class CountryInputTest < ActionView::TestCase
     assert_select %(select option[value="UA"] + #{COUNTRY_SELECT_SEPARATOR})
   end
 
+  test 'input generates a country select with a custom collection' do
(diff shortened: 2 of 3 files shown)


55728:

PR #1867, opened 2026-04-19
Fix label `for` not matching input `id` when collection value is nil

### LINKED ISSUE #1840: Label's for attribute doesn't match the id when nil is given in collection
## Environment

- Ruby **3.1.2p20**
- Rails **7.0.8**
- Simple Form **5.3.0**

## Current behavior

When a collection is explicitly given to radio-buttons and one of the values is `nil`, the generated `for` label for the `nil` one does not match its corresponding `id` with a difference of a single underscore character. For example,

```erb
<%= form.input :my_method, as: :radio_buttons, label: "ABC",
      collection: [['Yes', true], ['Undefined', nil]], checked: nil %>
```

for `Model` generates

```html
<fieldset class="form-group radio_buttons model_my_method">
<legend class="col-form-label pt-0">ABC</legend>
<input type="hidden" name="model[my_method]" value="" autocomplete="off" />
<!-- Valid one for "true" -->
<div class="form-check">
  <input class="form-check-input radio_buttons required" type="radio" value="true" name="model[my_method]"
    id="model_my_method_true" />
  <label class="form-check-label collection_radio_buttons"
    for="model_my_method_true">Yes</label>
</div>

<!-- Invalid one for "nil" -->
<div class="form-check">
  <input class="form-check-input radio_buttons required" type="radio" checked="checked" name="model[my_method]"
    id="model_my_method" />
  <label class="form-check-label collection_radio_buttons"
    for="model_my_method_">Undefined</label>
</div>
```

The second pair of `input` and `label` corresponds to the `nil`-parameter. As you see, the `for` has a trailing `_`, whereas that for `id` does not, and so they are not consistent.

## Expected behavior

The label and corresponding ID values must exactly agree.

55728:
Fixes #1840.

### Problem

When a collection contains a `nil` value, the generated label's `for` attribute does not match the corresponding input's `id`:

```erb
<%= form.input :my_method, as: :radio_buttons,
      collection: [['Yes', true], ['Undefined', nil]] %>
```

produces:

```html
<!-- ✅ "true" — matches -->
<input id="model_my_method_true" type="radio" value="true" />
<label for="model_my_method_true">Yes</label>

<!-- ❌ nil — trailing underscore mismatch -->
<input id="model_my_method" type="radio" value="" />
<label for="model_my_method_">Undefined</label>
```

Clicking the "Undefined" label does **not** select its radio button because the `for` / `id` values differ by a single trailing underscore.

### Cause

Rails' [`sanitize_attribute_name`](https://github.com/rails/rails/blob/db4bd562d90427c21a4525412c25372dd6994a65/actionview/lib/action_view/helpers/tags/collection_helpers.rb#L71-L73) unconditionally interpolates an underscore separator:

```ruby
def sanitize_attribute_name(value)
  "#{sanitized_method_name}_#{sanitized_value(value)}"
end
```

When `value` is `nil`, `sanitized_value(nil)` returns `""`, producing `"my_method_"` (trailing underscore). This value is passed to the label builder as the method name, which generates `for="model_my_method_"`.

Meanwhile, the radio button's `id` is generated by `add_default_name_and_id_for_value`, which has a **separate** nil-check that skips the value suffix entirely, producing `id="model_my_method"` (no trailing underscore).

### Fix

Override `sanitize_attribute_name` in `SimpleForm::Tags::CollectionExtensions` to skip the underscore when the sanitized value is empty:

```ruby
def sanitize_attribute_name(value)
  sanitized = sanitized_value(value)

  if sanitized.empty?
    sanitized_method_name.dup
  else
    "#{sanitized_method_name}_#{sanitized}"
  end
end
```

This aligns the label's `for` with the input's `id` for both `collection_radio_buttons` and `collection_check_boxes`, in both inline and nested boolean styles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Ruby 4.0 support (no changes required)
 * Support procs on validators for minlength/maxlength, and improve validators logic across the board to match Rails [#1859](https://github.com/heartcombo/simple_form/pull/1859)
+* Fix label `for` attribute not matching input `id` when collection contains a nil value [#1840](https://github.com/heartcombo/simple_form/issues/1840)
 
 ## 5.4.0
 
--- a/lib/simple_form/tags.rb
+++ b/lib/simple_form/tags.rb
@@ -26,6 +26,20 @@ module SimpleForm
         end.join.html_safe
       end
 
+      # Override Rails' sanitize_attribute_name to avoid generating a trailing
+      # underscore when the collection value is nil (e.g. "method_" instead of
+      # "method"), which causes the label's `for` attribute to not match the
+      # input's `id`. See https://github.com/heartcombo/simple_form/issues/1840
+      def sanitize_attribute_name(value)
+        sanitized = sanitized_value(value)
+
+        if sanitized.empty?
+          sanitized_method_name.dup
+        else
+          "#{sanitized_method_name}_#{sanitized}"
+        end
+      end
+
       def wrap_rendered_collection(collection)
         wrapper_tag = @options[:collection_wrapper_tag]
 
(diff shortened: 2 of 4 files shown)


55728:
Closing this as stale. The upstream bug was fixed in rails/rails#57209. Thanks!

PR #1868, opened 2026-04-20
Add `label_for` option to wrappers

### LINKED ISSUE #1800: Add "for" attribute in label wrapper
Trying to create a custom wrapper for checkboxes and radio buttons, I am hitting on a wall when trying to add the `for` attribute to a label wrapper.

The label contains markup with divs etcetera, which is allowed.

I tried this proposed solution, but being it rather old it does not seem to work:
https://github.com/heartcombo/simple_form/issues/869

The idea is to have a markup like this:

```
<label for="input_name" class="some clases">
    <div>Something</div>
    <div>
         <div>something else</div>
    </div>
</label>
```

Using the wrapper api, I do this:

```
a.wrapper tag: :label, class: "some classes" do |b|
    b.use :something, wrapped_with: {tag: :div}

    b.wrapper tag: :div do |c|
        c.use :something_else, wrapped_with: {tag: :div}
    end
end
```
The markup is  created, but there doesn't appear to be a method to add the `for="input_name"` in the label tag.

Is there any way to override this?
Shouldn't there be the possibility to use labels as wrappers, with their "for" attribute, for complex html markups?

Best regards,
Stavros

55728:
Closes #1800

When building custom wrappers with `tag: :label`, there was no way to automatically add the `for` attribute pointing to the corresponding input element.

This PR adds a `label_for` option to the wrapper DSL. When set to `true`, the wrapper resolves the input's `id` and adds it as the `for` attribute on the wrapping element.

### Usage

```ruby
config.wrappers :custom_label do |b|
  b.wrapper tag: :label, label_for: true, class: "custom-label" do |c|
    c.use :input
    c.wrapper tag: :div do |d|
      d.use :hint
    end
  end
end
```

This produces:

```html
<label for="user_name" class="custom-label">
  <input type="text" id="user_name" />
  <div>
    <span class="hint">...</span>
  </div>
</label>
```

### How it works

- `label_for: true` is opt-in — existing wrappers are not affected
- If `input_html: { id: "custom_id" }` is provided, that value is used
- Otherwise, `field_id` (Rails 7.0+) generates the standard input ID
- A `for` attribute set via `html: { for: "..." }` takes precedence (`||=`)
- Works correctly with nested forms (`simple_fields_for`) — `field_id` sanitizes the full `object_name` path consistently with Rails' input ID generation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+* Add `label_for` option to wrappers, allowing label wrapper tags to automatically include the `for` attribute pointing to the corresponding input. [#1868](https://github.com/heartcombo/simple_form/pull/1868)
+
 ## 5.4.1
 
 * Ruby 4.0 support (no changes required)
--- a/lib/simple_form/wrappers/many.rb
+++ b/lib/simple_form/wrappers/many.rb
@@ -59,11 +59,24 @@ module SimpleForm
         klass = html_classes(input, options)
         opts  = html_options(options)
         opts[:class] = (klass << opts[:class]).join(' ').strip unless klass.empty?
+
+        if @defaults[:label_for]
+          opts[:for] ||= resolve_label_for(input)
+        end
+
         input.template.content_tag(tag, content, opts)
       end
 
       def html_options(options)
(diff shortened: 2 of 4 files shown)


55728:

Issue #1216, opened 2015-03-06
HTML5 browser_validations is false by default not true


aliaksandrb:
Hi guys!
First of all thanks for the great gem, it makes development process much easier and faster! :+1: 

By the way, I've faced strange behavior recently and believe that it is a bug. Let me describe it.

I am using  simple_form for foundation via `rails generate simple_form:install --foundation`.

The documentation says :

> If you want to have all other HTML 5 features, such as the new field types, you can disable only the browser validation:
>   SimpleForm.browser_validations = false **# default is true**

But if we check it, than we see that  `SimpleForm.browser_validations = false` by default, not true

I found only two places where it could be changed:
[initializers/simple_form.rb](https://github.com/plataformatec/simple_form/blob/749c926140ee383e8249f15d1846f5287c5a4c46/lib/generators/simple_form/templates/config/initializers/simple_form.rb#L121)

``` ruby
config.browser_validations = false
```

and
[lib/simple_form.rb](https://github.com/plataformatec/simple_form/blob/749c926140ee383e8249f15d1846f5287c5a4c46/lib/simple_form.rb#L119)

``` ruby
mattr_accessor :browser_validations
@@browser_validations = true
```

And I suspect the problem is in the generator for `simple_for`installation
[install_generator.rb](https://github.com/plataformatec/simple_form/blob/749c926140ee383e8249f15d1846f5287c5a4c46/lib/generators/simple_form/install_generator.rb#L18)

It generates both files `simple_form.rb` and `simple_form_foundation.rb` (or bootstrap) even we chose only for Foundation.
Then Rails is loading initializers in alphabetical order, so browser_validations became false because of `initializers/simple_form.rb` and then this option used it foundation initializer because there is not override there.

So few questions:
- should the `browser_validations` be set to true in `initializers/simple_form.rb` (and both foundation and bootstrap initializers) ? - this is according to the README. Or update the README appropriately.
- what is the reason for generating `initializers/simple_form.rb` when we installing `simple_form` for Foundation/Bootstrap? Maybe it's fine to not doing it and duplicate common options in each initializer if any?
- Want I send an PR ? :blush:

jedschneider:
@aliaksandrb does this seem like a similar issue? https://github.com/plataformatec/simple_form/issues/1178 I have a PR I'm working on, but can't seem to get support for a valid test case.

aliaksandrb:
[...]

Issue #319, opened 2011-09-11
disable browser validation by default


rafaelfranca:
There are many bug reports and emails to list about errors because the browser validations is enabled.

PR #1869, opened 2026-04-20
Fix misleading comment in README about browser_validations default

### LINKED ISSUE #1746: README.md says browser_validations = true by default, but generator makes it false
Similar to #1216 which was closed

## Environment

- Ruby **2.7.3**
- Rails **6.1.3.2**
- Simple Form **5.1.0**

## Current behavior

`rails generate simple_form:install`
Creates `config/initializers/simple_form.rb` where line 130 is:

```ruby
config.browser_validations = false
```

## Expected behavior

The README.md says that it is **true** by default
https://github.com/heartcombo/simple_form/blame/master/README.md#L1111-L1113

Should the generator not create it as `true` if the readme says `true` is the default?

55728:
The README contains a code example with a comment stating `# default is true`:

```ruby
SimpleForm.browser_validations = false # default is true
```

While `true` is the default in `lib/simple_form.rb`, the generator intentionally sets it to `false` in the initializer (see #319). This has caused recurring confusion, as reported in #1216 and #1746.

This PR removes the misleading inline comment so that users are not led to believe their app is running with `browser_validations = true` after running the generator.

Fixes #1746

### DIFF
--- a/README.md
+++ b/README.md
@@ -1170,7 +1170,7 @@ If you want to have all other HTML 5 features, such as the new field types, you 
 the browser validation:
 
 ```ruby
-SimpleForm.browser_validations = false # default is true
+SimpleForm.browser_validations = false
 ```
 
 This option adds a new `novalidate` property to the form, instructing it to skip all HTML 5


55728:

PR #1870, opened 2026-04-20
Fix I18n lookup when child_index contains non-word characters

### LINKED ISSUE #1795: I18n lookup breaks when using child_index with non-word characters
## Environment

- Ruby **2.7.6**
- Rails **6.1.7**
- Simple Form **5.1.0**

## Current behavior

In [form_builder.rb](/heartcombo/simple_form/blob/main/lib/simple_form/form_builder.rb#L466) there's the following code:

```ruby
    def lookup_model_names #:nodoc:
      @lookup_model_names ||= begin
        child_index = options[:child_index]
        names = object_name.to_s.scan(/(?!\d)\w+/).flatten
        names.delete(child_index) if child_index
        names.each { |name| name.gsub!('_attributes', '') }
        names.freeze
      end
    end
```

I was using `"#index"` as the `child_index` option, to distinguish it from normal names (I'm doing some javascript fiddling to replace the `"#index"` with something proper later on). 

I noticed the hints weren't being properly translated. Looking through the old issues I found #245 and #246, which looked similar to my issue. I realized it was because the regex `/(?!\d)\w+/` would only match word-characters, so it didn't match the `#`-character.

I monkey-patched the regex to `/(?!\d)[\w#]+/` or even `/(?!\d)[^\[\]]+/` which fixed my issue.

Not sure if SimpleForm should be more tolerant of special characters in `child_index` or not, but I just wanted to raise it in case other people encountered this problem.

55728:
Closes #1795

## Problem

`lookup_model_names` uses the regex `/(?!\d)\w+/` to scan `object_name` into segments, then attempts to remove `child_index` from the result via `names.delete(child_index)`.

When `child_index` contains non-word characters (e.g. `"#index"`), the regex never captures it as a single segment, so `delete` silently does nothing. This leaves stale segments in the lookup names, breaking I18n resolution for hints, labels, and other translated components in nested forms.

## Solution

Remove `child_index` (with its surrounding brackets) from the `object_name` string **before** the regex scan, rather than trying to delete it from the results afterward. This handles any `child_index` value regardless of the characters it contains.

## Verification

- Added a test for `child_index: "#index"` (non-word character case)
- Full test suite passes (807 runs, 0 failures)

### DIFF
--- a/lib/simple_form/form_builder.rb
+++ b/lib/simple_form/form_builder.rb
@@ -473,8 +473,9 @@ module SimpleForm
     def lookup_model_names #:nodoc:
       @lookup_model_names ||= begin
         child_index = options[:child_index]
-        names = object_name.to_s.scan(/(?!\d)\w+/).flatten
-        names.delete(child_index) if child_index
+        name = object_name.to_s
+        name = name.gsub("[#{child_index}]", '') if child_index
+        names = name.scan(/(?!\d)\w+/).flatten
         names.each { |name| name.gsub!('_attributes', '') }
         names.freeze
       end
--- a/test/components/label_test.rb
+++ b/test/components/label_test.rb
@@ -165,6 +165,25 @@ class IsolatedLabelTest < ActionView::TestCase
     end
   end
 
+  test 'label does correct i18n lookup for nested has_many models when child_index contains non-word characters' do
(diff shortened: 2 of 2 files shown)


55728:

